PR #3, opened 2016-11-04
PTNFLY 1868 Setup a perf test harness

priley86:
* adds [sitespeed.io](https://www.sitespeed.io/) automated perf testing
* adds @bleathem's pf-tabs perf tests 
* refactors webpack to use webpack config/ multiple entries allow component separation
* adds `index.js` for each component and individual component `package.json` files
* adds [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer)

webpack-bundle-analyzer

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -1,12 +1,18 @@
 var gulp = require('gulp'),
   browserSync = require('browser-sync').create(),
+  browserSyncPort = 3000,
+  baseUrl = 'http://localhost:' + browserSyncPort,
   eslint = require('gulp-eslint'),
   ignore = require('gulp-ignore'),
   karma = require('karma').Server,
   path = require('path'),
   sass = require('gulp-sass'),
   $ = require('gulp-load-plugins')(),
-  webpack = require('webpack-stream');
+  webpack = require('webpack-stream'),
+  runSequence = require('run-sequence'),
+  sitespeedio = require('gulp-sitespeedio'),
+  runBundleAnalyzer = false,
+  BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
 
 gulp.task('font', function(){
   return gulp.src([
@@ -54,18 +60,46 @@ gulp.task('test-debug', function (done) {
   }, done).start();
 });
 
+gulp.task('sitespeedio', ['serve'], function(done){
+  var run = sitespeedio({
+    urls: [
+      baseUrl + '/app/app.html?file=pf-alert',
+      baseUrl + '/app/app.html?file=pf-tabs',
+      baseUrl + '/app/app.html?file=pf-utilization-bar-chart',
+      baseUrl + '/app/app.html?file=pf-list-view'
+    ],
+    browser: 'chrome',
+    resultBaseDir: './perf/',
+    connection: 'cable',
+    html: true,
+    budget: {
+      page: {
+        pageWeight:1000000,
+        imageWeight: 300000,
+        jsWeight: 500000,
+        cssWeight: 500000
+      }
+    }
+  });
+  run(done);
+});
+
+gulp.task('perf', function(done){
+  runSequence('build', 'sitespeedio', done)
+});
+
+gulp.task('bundleAnalyzer', function(done){
+  runBundleAnalyzer = true;
+  runSequence('build', done)
+});
+
 gulp.task('webpack', ['js'], function() {
-  return gulp.src(['dist/es2015/*/*.js'])
-    .pipe(webpack({
-      resolve: {
-        root: [
-          path.join(__dirname, "dist/es2015/pf-alert"),
-          path.join(__dirname, "dist/es2015/pf-tabs"),
-          path.join(__dirname, "dist/es2015/pf-utilization-bar-chart"),
-          path.join(__dirname, "dist/es2015/pf-utils")
-      ]}
-    }))
-    .pipe($.rename('patternfly.js'))
+  var webpackConfig = require('./webpack.config.js');
+  if(runBundleAnalyzer){
+    webpackConfig.plugins.push(new BundleAnalyzerPlugin());
+  }
+  return gulp.src('src/patternfly.js')
+    .pipe(webpack( webpackConfig ))
     .pipe(gulp.dest('dist/js'));
 });
 
@@ -75,7 +109,8 @@ gulp.task('serve', function() {
   browserSync.init({
     server: {
       baseDir: './'
-    }
+    },
+    port: browserSyncPort
   });
 
   gulp.watch('index.html', ['build']);
--- a/src/patternfly.js
+++ b/src/patternfly.js
@@ -1,0 +1,14 @@
+/** PF Alert Component **/
+require('pf-alert.component.js');
+
+/** PfListView Component **/
+require('pf-list-view.component.js');
+
+/** PF Tabs Component **/
+require('pf-tabs.component');
+
+/** PF Utilization Bar Chart **/
+require('pf-utilization-bar-chart.component');
+
+/** PF Utils **/
+require('pf-utils');
--- a/src/pf-tabs/index.js
+++ b/src/pf-tabs/index.js
@@ -1,0 +1,2 @@
+/** PF Tabs Component **/
+require('pf-tabs.component');
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,0 +1,52 @@
+var webpack = require('webpack');
+var path = require('path');
+
+module.exports = {
+
+  context: __dirname,
+
+  entry: {
+    'patternfly': './src/patternfly',
+    'pf-alert': './src/pf-alert/index',
+    'pf-list-view': './src/pf-list-view/index',
+    'pf-tabs': './src/pf-tabs/index',
+    'pf-utilization-bar-chart': './src/pf-utilization-bar-chart/index',
+    'pf-utils': './src/pf-utils/index'
+  },
+
+  resolve: {
+    root: [
+      path.join(__dirname, "src/pf-alert"),
+      path.join(__dirname, "src/pf-list-view"),
+      path.join(__dirname, "src/pf-tabs"),
+      path.join(__dirname, "src/pf-utilization-bar-chart"),
+      path.join(__dirname, "src/pf-utils")
+    ]
+  },
+
+  output: {
+    path: path.join(__dirname, './dist/js/'),
+    publicPath: './',
+    filename: '[name].js'
+  },
+
+  externals: {
+  },
+
+  plugins: [
+  ],
+
+  module: {
+    loaders: [
+      //js loader
+      {
+        loader: "babel",
+
+        // Options to configure babel with
+        query: {
+          presets: ['es2015']
+        }
+      }
+    ]
+  }
+};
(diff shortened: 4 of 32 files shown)


gulpfile.js
```diff
@@ -54,18 +58,36 @@ gulp.task('test-debug', function (done) {
...
+gulp.task('perf', function(done){
+  runSequence('build', 'sitespeedio', done)
+});
+
 gulp.task('webpack', ['js'], function() {
-  return gulp.src(['dist/es2015/*/*.js'])
-    .pipe(webpack({
-      resolve: {
-        root: [
-          path.join(__dirname, "dist/es2015/pf-alert"),
```
dlabrecq: What I was trying to achieve here was to eliminate the relative paths. If this component is required elsewhere (e.g., in another repo), it won't build unless the exact directory structure is used.

src/patternfly.js
```diff
@@ -0,0 +1,17 @@
+/** PF Alert Component **/
+require('./pf-alert/pf-alert.component.js');
```
dlabrecq: I like the external file, but I was letting the components require the modules needed. Do we really need to include the template here, for example?

src/pf-alert/pf-alert.component.js
```diff
@@ -1,5 +1,5 @@
-import {default as tmpl} from 'pf-alert.template';
-import {default as utils} from 'pf-utils';
+import {default as tmpl} from './pf-alert.template';
+import {default as utils} from '../pf-utils/pf-utils';
```
dlabrecq: This is what I'm trying to avoid. I don't believe we should be hard coding relative paths here. Like I tried to suggest above, if another component requires pf-alert (e.g., in another repo), this won't build unless the directory structure is exactly the same.

src/pf-tabs/index.js
```diff
@@ -0,0 +1,5 @@
+/** PF Tabs Component **/
+require('./pf-tab.component');
+require('./pf-tab.template');
+require('./pf-tabs.component');
+require('./pf-tabs.template');
```
dlabrecq: I don't understand why this is required considering tab and tabs import their own dependencies?

webpack.config.js
```diff
@@ -0,0 +1,41 @@
...
+  module: {
+    loaders: [
+      //js loader
+      {
+        loader: "babel",
+
+        // Options to configure babel with
+        query: {
+          presets: ['es2015']
+        }
```
dlabrecq: It appears that the "js" target already runs babel and outputs to the dist/es2015 directory. However, I see babel here as well? And I don't see dist/2015 as the source, but the new index.js files.

The reason I had the dist/2015 directory was for running webpack in your own repo, if necessary. When we had more seperate repos, it wasn't a good idea to webpack something that has already been webpack'ed. Thus, the dist/2015 directory.

priley86:
I agree and disagree :).

The core argument I am making here is that we want to expose both independently loadable scripts (babel/webpack compiled es5 scripts) AND es6 modules/imports. This allows downstreams who **don't** use webpack to simply load a script tag (as in the perf example test I simply load pf-alerts.js without worrying about webpack at all). Providing an es5 prepackaged component makes it super easy for those who just want to drop a script reference in their legacy page (or Polymer component) to get up and running quickly and move on with their day.

For those who **do** use webpack/es6 module loaders to import a webpack module in their component, I've written an example of what an import might look like in my React app here: https://github.com/priley86/patternfly-react-starter/commit/c704ca2aeab6112ed7289b3c48de40b2590f45f5

One thing to note on this though is I had to give a full path in my import statement (not ideal either). There are four ways I know of handling this. Let me give an thorough description of each.

1. Split up components in separate repositories/npm packages and specify the es6 module in your package.json's `main` attribute (though we've already ruled this option out). This is very common in the React/Web Component community though and it's what allows you to just say `import React from 'react'; ` for example. This would pull in all modules in the React es6 distributable. 

2. Specify a specific module name in the import, i.e. `import { PropTypes } from 'react';` or in our case `import {PfAlert} from 'patternfly-webcomponents';`. This references a specific es6 module for the scope it is imported in, but you still need to load the entire es6 bundle into your build process. If your goal is to distribute all of these modules together (which it is in the React case), that's fine, but I don't think it's **best** case. Webpack2/tree-shaking allows your build to reduce this (thus resolving this), but I'd prefer to not assume every downstream will have this if we don't have to. This is the Onsen.io/Angular2/React approach FWIW. Example here:
(one react-onsenui.js distributable containing all modules):
https://github.com/OnsenUI/react-onsenui/tree/master/dist
(importing a specific page module in a downstream app):
https://github.com/argelius/react-onsenui-redux-weather/blob/master/components/MainPage.js

3. Create a separate webpack "entry" for each component script (which creates a minimal module containing only the referenced pf-alert component scripts). This was the option I went with here - you can see the separate bundles exist and can be imported downstream individually. It's not ideal that I need to give an absolute reference (node_modules/...), but it constrains my import to the smallest bundle possible, and I don't depend on Webpack2 (I could use rollup or require or common or something else). Using multiple webpack entries (also called code splitting) is an advanced approach that is starting to become common in the PWA scene and server rendered React apps. You can read more about this [here](https://medium.com/@addyosmani/progressive-web-apps-with-react-js-part-2-page-load-performance-33b932d97cf2#.fzcmigsjh) if interested. 
 
 4. Reference all of the component scripts directly from their es6 `src`, i.e. `import '../../node_modules/patternfly-webcomponents/src/pf-alert/pf-alert.component.js' , `import '../../node_modules/patternfly-webcomponents/src/pf-alert/pf-alert.template.js' etc etc. This is a lot more typing downstream, but it works and provides direct access to every component script if desired.

So as you can see we are in a bit of a pickle - as we'd like to expose individual component scripts without forcing downstreams to load the monolithic bundle or assume they're using tree-shaking/webpack2, and still support those that don't use webpack/es6 at all. For this reason, I'll propose that we go with Option 3 with the addition of marking "patternfly.js" as our "main" in `package.json`. This way we get the syntax of Option 2 (for those who load all modules and may choose to tree-shake), and we can still support Option 3 performance goals without webpack2. If we decide that it is possible to publish these components individually or eventually switch to Option 1 long term, it's not a huge change downstream either...

Let me know if this makes sense and I'll be happy to add the "main" attribute. It may be helpful to play around with some downstream webpack apps and see if you still disagree...

src/patternfly.js
priley86: Relative paths should work fine here and downstream (although I realize we had a different goals originally with separate repos). This is common in the Onsen UI project and the Angular2/React projects. Ex:
https://github.com/OnsenUI/OnsenUI/blob/master/core/src/elements/ons-alert-dialog/index.js#L18

dlabrecq:
I don't necessarily disagree with having independently loadable scripts. It's the implementation that I'm having trouble understanding. Maybe just some clarification is all that's needed here?

What I feel strongly about is reintroducing relative paths -- this is a red flag to me. I can tell you relative paths not work well in a separate project/repo, unless you can somehow guarantee the same directory structure? (Maybe it worked for your prototype, but it didn't for me.) Thus, I find it difficult to see how that "should work fine downstream". I immediately ran into this problem as soon as I imported a web component. I found multiple blogs stating not to use relative paths -- I don't want to dismiss that warning.

What is not clear about the implementation is having all component files defined in an index.js file? Why all files? This doesn't seem correct to me because it defeats the purpose of defining dependencies via module imports/exports. For example, what purpose does it sever to "require" the templates in index.js when the module imports already do that? And if do that, why do we even need modules if we have a file that requires everything?

The other thing that's not clear is related to the "js" target. This target is already running Babel, but so is webpack. Why both?

That said, perhaps the perf test should be a separate PR? Right now, I feel there are too many cooks in the kitchen. I get it at a high level, but discussing offline would help clarify what your doing with the implementation?

dlabrecq:
I understand you're trying to rebuild webpack downsteam, but the fundamental difference is that we're using ES6 modules and we must allow them to be imported downstream. When you add a webcomponent as a dependency, via an ES6 import, webpack must be able to locate it. And unless both upstream and dowstream share the same directory structure, webpack cannot locate the dependency and builds fail. I believe Onsen/Angular 2 use HTML imports, so following what they did may not be the right choice with ES6 modules? Regardless, lets meet to discuss.

gulpfile.js
priley86: @dlabrecq i've added `root` to the webpack config file so this behavior stays in tact in the `src` and `dist` folder. FWIW I don't think it bad to have relative path resolves in your local repo (just when consuming downstream node_modules).  It would save us from having to keep appending to this configuration, but I'll follow the pattern if we think it useful.

webpack.config.js
priley86: @dlabrecq we're "webpacking" all of the distributables now in this PR (i've just changed the js target to use a webpack config file). This should give adequate usage downstream in both es5 and es6 contexts.

priley86:
@dlabrecq i've updated this PR to reflect Option 1 (introducing nested package.jsons), and I tried to respect our preference to not use relative path in the `src`. Although this makes the `import` syntax shorter, it requires us to add more configurations in webpack, i.e.: 
```
  resolve: {
    root: [
      path.join(__dirname, "src/pf-alert"),
      path.join(__dirname, "src/pf-tabs"),
      path.join(__dirname, "src/pf-utilization-bar-chart"),
      path.join(__dirname, "src/pf-utils")
    ]
  },
```
Relative imports are still common in other projects, but I'm OK with this if it's still intended.  

Introducing nested packages still gets us in the mindset of module separation and `index.js` allows us to combine as many scripts as needed (for example, the pf-tabs components needs both `pf-tab` and `pf-tabs`). It's also default on `npm init` and a community standard...

dlabrecq:
Thanks for removing the relative path.

I downloaded the changes and started testing the index.js files. I can remove the templates for example and webpack will still pull in the required module dependencies as expected. Obviously, this is because they are imported by the web component. For example, I can reduce the index.js file for tabs from:

require('pf-tab.component');
require('pf-tab.template');
require('pf-tabs.component');
require('pf-tabs.template');

to:

require('pf-tab.component');
require('pf-tabs.component');

I can even go so far as to remove the require statement for 'pf-tab.component'. I still see the templates being built and webpack includes everything as expected. In fact, I still see the templates being added to the dist/es2015/pf-tabs directory.

So, why do we want index.js file to list all these files? Isn't this circumventing the ES6 module imports defined at the web component level?

Can we just list the main file of the web component and let webpack pull in the dependencies defined by the ES6 imports?

I'm also still curious about Babel being the webpack loader. It's not clear if we're running Babel twice?

priley86:
@dlabrecq i've added [webpack-bundle-analyzer](https://www.npmjs.com/package/webpack-bundle-analyzer) to this PR as well. This should help clarify the different bundles generated and provide a thorough analysis of their makeup and size. Also made the pf-list-view consistent with the other changes...

i've removed the additional explicit references in `index.js` and agree that these can be handled naturally via es6/imports (as you've noted). It's less obvious what components scripts are pulled in, but I'm OK with that (as it can easily be visualized now).  

Regarding the Babel loader, the `dist/es2015` folder and `js` gulp task simply provide es6 transpiled es5 scripts (via Babel), however these scripts are not *standalone* like the ones in `dist/js` as they do not have a module loader (such as the webpack module loader written in `dist/js`), so that would still be necessary downstream. It is unlikely a downstream would import the transpiled es5 rather than es6 from `src` into their build process, but if desired, it would work.

dlabrecq:
Thanks Patrick. Glad we nailed all this down early.

PR #6, opened 2016-11-19
Initial i18n implementation.

dlabrecq:
Made changes per code review with @bleathem. The pf-i18n tag now accepts an object literal containing simple name value pairs. It also accepts an object with a getMsg function to retrieve translated messages.

Note that the build creates a .pot file containing all message strings to be translated. However, we decided not to provide the translated .po files at this time.


priley86:
@dlabrecq fantastic! I like this approach a lot (and will enable us to extend i18n as needed). Will make a few nit comments, but overall, happy to see this merged! 
👍 👍

app/i18n/de-DE/patternfly.json
```diff
@@ -0,0 +1,15 @@
...
+   "domain": "messages",
+   "locale_data": {
+      "messages": {
+         "": {
+            "domain": "messages",
+            "plural_forms": "nplurals=2; plural=(n != 1)",
+            "lang": "de-DE"
+         },
+         "Hello World!": [
+            "Hello World! (de-DE)"
```
priley86: Is the translation here `Hallo Welt` ? Should we spell it out?

app/i18n/i18n.js
```diff
@@ -0,0 +1,95 @@
...
+    this._fetchLocaleData = function(url) {
+      let xmlhttp = new XMLHttpRequest();
+      xmlhttp.onreadystatechange = function () {
+        if (this.readyState === 4 && this.status === 200) {
+          const localeData = JSON.parse(this.responseText);
+          self._setLocaleData(localeData);
+        }
+      };
+      xmlhttp.open("GET", url, false);
+      xmlhttp.send();
```
priley86: I think a lot of frameworks are making use of the `fetch` polyfill/implementation now (see [here](https://github.com/github/fetch)), but happy to accept the more vanilla/cross supported use of `xmlHttp`. Great stuff.

priley86:
@dlabrecq i will start trying to implement this in downstream React/WC implementations once merged. 

bitmoji

app/i18n/i18n.js
dlabrecq: I'm all for anything that makes things easier for us. However, this is just the example app. I could have easily used jQuery, but tried to keep things simple.

Cool idea, tho. Maybe we can use that polyfill somewhere in our utils?

app/i18n/fr/patternfly.json
```diff
@@ -0,0 +1,15 @@
...
+   "domain": "messages",
+   "locale_data": {
+      "messages": {
+         "": {
+            "domain": "messages",
+            "plural_forms": "nplurals=2; plural=(n > 1)",
+            "lang": "fr"
+         },
+         "Hello World!": [
+            "Hello World! (fr)"
```
dlabrecq: Ok, I'll add these translations.

PR #14, opened 2016-12-15
introduce pf-tooltip

priley86:
* introduces a new pf-tooltip web component based on the [vanilla bootstrap tooltip](https://github.com/thednp/bootstrap.native/blob/master/lib/tooltip-native.js)

This is downstream work and not part of the current sprint. No rush on this review ;)

### DIFF
--- a/src/pf-tooltip/index.html
+++ b/src/pf-tooltip/index.html
@@ -1,0 +1,48 @@
+<!doctype html>
+<html>
+<head>
+    <title>pf-tooltip example</title>
+    <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/patternfly/3.13.0/css/patternfly-additions.css">
+    <link rel="stylesheet" href="/dist/css/patternfly.css">
+    <script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.22/webcomponents.js"></script>
+    <script src="/dist/js/pf-utils.js"></script>
+    <script src="/dist/js/pf-tooltip.js"></script>
+</head>
+<body>
+<div class="container">
+    <div class="page-header">
+        <h1>Tooltips</h1>
+    </div>
+    <hr>
+
+    <pf-tooltip id="tooltip-left" placement="left" targetSelector="#btn-left">
+        Tooltip text on Left
+    </pf-tooltip>
+    <button id="btn-left" type="button" class="btn btn-primary" aria-describedby="tooltip-left">Left</button>
+
+    <pf-tooltip id="tooltip-top" placement="top" targetSelector="#btn-top">
+        <span>Tooltip on top</span>
+    </pf-tooltip>
+    <button id="btn-top" type="button" class="btn btn-secondary" aria-describedby="tooltip-top">Top</button>
+
+    <pf-tooltip id="tooltip-bottom" placement="bottom" targetSelector="#btn-bottom">
+        <span>Tooltip on bottom</span>
+    </pf-tooltip>
+    <button id="btn-bottom" type="button" class="btn btn-warning" aria-describedby="tooltip-bottom">Bottom</button>
+
+    <pf-tooltip id="tooltip-right" placement="right" targetSelector="#btn-right">
+        <span>Tooltip on right</span>
+    </pf-tooltip>
+    <button id="btn-right" type="button" class="btn btn-danger" aria-describedby="tooltip-right">Right</button>
+
+    <br/>
+    <br/>
+    <script>
+        setTimeout(function(){
+          var pfTooltipRight = document.querySelector('[targetSelector="#btn-right"]');
+          pfTooltipRight.setInnerHtml("<span>Right content changed!</span>");
+        });
+    </script>
+</div>
+</body>
+</html>
--- a/src/pf-tooltip/pf-tooltip.component.js
+++ b/src/pf-tooltip/pf-tooltip.component.js
@@ -1,0 +1,327 @@
+import {default as tmpl} from 'pf-tooltip.template';
+import {pfUtil} from 'pf-utils.js';
+
+/**
+ * <b>&lt;pf-tooltip&gt;</b> element for Patternfly Web Components
+ *
+ * @example {@lang xml}
+ * <pf-tooltip animation="fade" targetSelector="#btn-left" placement="left" delay="100" duration="150" containerSelector="#container"></pf-alert>
+ *
+ * @prop {string} animation the animation class
+ * @prop {string} targetSelector the target element selector
+ * @prop {string} placement left, right, top, bottom
+ * @prop {number} delay animation delay (ms)
+ * @prop {number} duration animation duration (ms)
+ * @prop {string} containerSelector the container element selector
+ */
+
+export class PfTooltip extends HTMLElement {
+
+  /**
+   * Reinitializes tooltip component with attribute values and resets content
+   */
+  init () {
+    this.element = this;
+    this.content = this._innerHtml || this.element.innerHTML;
+    this.tooltip = null;
+    this._targetSelector = this.getAttribute('targetSelector');
+    this._target = this._targetSelector ? document.querySelector(this._targetSelector) : this;
+    this._animation = this.getAttribute('animation') ? this.getAttribute('animation') : 'fade';
+    this._placement = this.getAttribute('placement') ? this.getAttribute('placement') : 'right';
+    this._delay = parseInt(this.getAttribute('delay')) || 100;
+    this._mouseHover = ('onmouseleave' in document) ? [ 'mouseenter', 'mouseleave'] : [ 'mouseover', 'mouseout' ];
+    this._tipPositions = /\b(top|bottom|left|top)+/;
+    this._duration = pfUtil.isMSIE && pfUtil.isMSIE < 10 ? 0 : (parseInt(this.getAttribute('duration')) || 150);
+    this._containerSelector = this.getAttribute('containerSelector');
+    this._container = this._containerSelector ? document.querySelector(this._containerSelector) : document.body;
+
+    if (this._target) {
+      //create open event listeners
+      this._target.addEventListener(this._mouseHover[0], (e) => {
+        this.open(e);
+      }, false);
+
+      //create close event listener
+      this._target.addEventListener(this._mouseHover[1], (e) => {
+        this.close(e);
+      }, false);
+    }
+  }
+
+  /**
+   * Called when an instance was inserted into the document
+   */
+  attachedCallback () {
+    this.init();
+
+    //handleContentChanged
+    this.element.addEventListener('handleContentChanged', (e) => {
+      this.init();
+    }, false);
+  }
+
+  /**
+   * Called when element's attribute value has changed
+   *
+   * @param {string} attrName The attribute name that has changed
+   * @param {string} oldValue The old attribute value
+   * @param {string} newValue The new attribute value
+   */
+  attributeChangedCallback (attrName, oldValue, newValue) {
+    this.init();
+  }
+
+  /**
+   * Called when an instance of the element is created
+   */
+  createdCallback () {
+    this._template = document.createElement('template');
+    this._template.innerHTML = tmpl;
+    this._timer = 0;
+  }
+
+  /**
+   * Sets tooltip the inner HTML
+   * @param {string} html string
+   */
+  setInnerHtml(html) {
+    this._innerHtml = html;
+    this.element.dispatchEvent(new CustomEvent('handleContentChanged', {}));
+  }
+
+  /**
+   * Get the animation class
+   *
+   * @returns {string} The animation class
+   */
+  get animation () {
+    return this._animation;
+  }
+
+  /**
+   * Set animation class
+   *
+   * @param {string} value The animation class
+   */
+  set animation (value) {
+    if (this._animation !== value) {
+      this._animation = value;
+      this.setAttribute('animation', value);
+    }
+  }
+
+  /**
+   * Get the tooltip containerSelector
+   *
+   * @returns {string} The container element selector
+   */
+  get containerSelector () {
+    return this._containerSelector;
+  }
+
+  /**
+   * Set the tooltip containerSelector
+   *
+   * @param {string} value The container element selector
+   */
+  set containerSelector (value) {
+    if (this._containerSelector !== value) {
+      this._containerSelector = value;
+      this._container = document.querySelector(this._containerSelector);
+      this.setAttribute('containerSelector', value);
+    }
+  }
+
+  /**
+   * Get the animation duration
+   *
+   * @returns {string} The animation duration
+   */
+  get duration () {
+    return this._duration;
+  }
+
+  /**
+   * Set the animation duration
+   *
+   * @param {string} value The animation duration
+   */
+  set duration (value) {
+    if (this._duration !== value) {
+      this._duration = value;
+      this.setAttribute('duration', value);
+    }
+  }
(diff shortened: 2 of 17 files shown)


src/pf-utils/pf-utils.js
```diff
@@ -38,4 +38,76 @@ let I18nUtil = function() {
...
+ * --------------------------------------------------------------------------
+ * PfUtil
+ * Native Javascript for Bootstrap 3 | Internal Utility Functions
+ * --------------------------------------------------------------------------
+ */
+
+class PfUtil {
+
+  constructor () {
+    this.isIE = (new RegExp("MSIE ([0-9]{1,}[\.0-9]{0,})").exec(navigator.userAgent) !== null) ? parseFloat( RegExp.$1 ) : false;
```
dlabrecq: This won't detect all versions of IE (e.g., Edge). We're using this in the table JS.

return /(MSIE|Trident\/|Edge\/)/i.test(window.navigator.userAgent);

src/pf-utils/pf-utils.js
```diff
@@ -38,4 +38,76 @@ let I18nUtil = function() {
...
+export {i18n};
+
+/**
+ * --------------------------------------------------------------------------
+ * PfUtil
+ * Native Javascript for Bootstrap 3 | Internal Utility Functions
+ * --------------------------------------------------------------------------
+ */
+
+class PfUtil {
```
dlabrecq: Can this be renamed something more specific (e.g., NativeUtil or NativeBootstrapUtil)? (Not certain this code is dependent on Bootstrap?) The intention was to add specific objects like i18nUtil, but not PfUtil itself which kind of becomes a catch-all for random functions.

Alternatively, we might want to move each object to its own file?

src/pf-tooltip/pf-tooltip.component.js
```diff
@@ -0,0 +1,304 @@
...
+    let tooltipInner = clone.querySelector('.tooltip-inner');
+
+    //set tooltip content
+    tooltipInner.innerHTML = this.content;
+
+    //append to the container
+    this.container.appendChild(clone);
+
+    //set reference to appended node
+    let tooltips = this.container.querySelectorAll('.tooltip');
```
dlabrecq: This selector name (.tooltip) may be too generic? It's possible that this could match something else in the page. That is, considering container could be the body tag.

The '.tooltip-inner' selector looks ok because it's found under this._template.content.

src/pf-tooltip/pf-tooltip.component.js
```diff
@@ -0,0 +1,304 @@
+import {default as tmpl} from 'pf-tooltip.template';
+import {pfUtil} from 'pf-utils';
+
+/**
+ * <b>&lt;pf-tooltip&gt;</b> element for Patternfly Web Components
+ *
+ * @example {@lang xml}
+ * <pf-tooltip animation="fade" target="#btn-left" placement="left" delay="100" duration="150" container="#container"></pf-alert>
```
dlabrecq: Curious why the container must be specified. I understand the default is the body tag; however, couldn't we use the parent of the given target or the pf-tooltip element itself?

src/pf-tooltip/pf-tooltip.component.js
```diff
@@ -0,0 +1,304 @@
...
+export class PfTooltip extends HTMLElement {
+
+  /**
+   * Reinitializes tooltip component with attribute values and resets content
+   */
+  init () {
+    this.element = this;
+    this.content = this.element.innerHTML;
+    this.tooltip = null;
+    this._target = this.getAttribute('target') ? document.querySelector(this.getAttribute('target')) : this;
```
dlabrecq: I'm thinking we should use the attribute name targetSelector? Looking at the attribute only, it's not clear that the target accepts an ID, class, etc.

I see "selector" mentioned in the JsDoc, but making the attribute names clear would be helpful. The datatables.net attributes append "Selector" to the attributes, for example. 

It would help folks understand that they need to use "#btn-left" instead of "btn-left", for example.

src/pf-utils/pf-utils.js
```diff
@@ -38,4 +38,76 @@ let I18nUtil = function() {
...
+
+  constructor () {
+    this.isIE = (new RegExp("MSIE ([0-9]{1,}[\.0-9]{0,})").exec(navigator.userAgent) !== null) ? parseFloat( RegExp.$1 ) : false;
+    this.mouseHover = ('onmouseleave' in document) ? [ 'mouseenter', 'mouseleave'] : [ 'mouseover', 'mouseout' ];
+    this.tipPositions = /\b(top|bottom|left|top)+/;
```
dlabrecq: A mouseHover variable doesn't seem generic enough for a util object that could potentially be used by other components. This seems like something this belongs in the component istelf?

Likewise, tipPositions seems very specific to the tooltip component. I'm don't know that this would be useful to other components?

src/pf-tooltip/index.html
```diff
@@ -0,0 +1,51 @@
...
+
+    <br/>
+    <br/>
+    <script>
+        setTimeout(function(){
+          var pfTooltipRight = document.querySelector('[target="#btn-right"]');
+          pfTooltipRight.innerHTML = "Right content changed!";
+
+          var event = new CustomEvent('handleContentChanged',{});
+          pfTooltipRight.dispatchEvent(event);
```
dlabrecq: Does the app developer need to dispatch this custom event or is this just a test?

Wondering if we should consider introducing something like js-signals, which would make it a lot easier to publish and subscribe to custom events like this?

https://millermedeiros.github.io/js-signals/

src/pf-tooltip/pf-tooltip.component.js
```diff
@@ -0,0 +1,304 @@
...
+   */
+  removeTooltip () {
+    this.tooltip && this.container.removeChild(this.tooltip);
+    this.tooltip = null;
+  }
+
+  /**
+   * Creates the tooltip
+   */
+  createTooltip () {
```
dlabrecq: Should createTooltip be private (i.e., with an underscore)?

Likewise, removeTooltip, open, close, etc. seem like they could possibly be private functions?

priley86:
thanks for the feedback @dlabrecq! I certainly anticipated quite a bit of review was still needed on this. I will try to respond and make updates next week...

src/pf-utils/pf-utils.js
priley86: +1 for moving each related Util to its own file. I'm not entirely sold on naming it NativeBootstrapUtil as we'll be changing this quite a bit, and likely appending to it. I would think that a large number of common routines that used to be in jQuery (like addClass, removeClass, getClosest) now being implemented in Vanilla js would fit into the same logical utility file (and I guess my preference is just PfUtils for that). Let me know if this works and I'm happy to split it out.

src/pf-utils/pf-utils.js
priley86: +1. Will update for Trident / Edge.

src/pf-utils/pf-utils.js
priley86: Fine with moving these mouseHover and tip positions into the pf-tooltip component and out of PFUtils.

src/pf-tooltip/pf-tooltip.component.js
priley86: I like having container as an attribute for the added flexibility there, but am fine with changing the default container to the parent element (instead of body). Was just following the BootstrapNative pattern with that...

src/pf-tooltip/pf-tooltip.component.js
priley86: Yep..i'm fine with `targetSelector` ... i've seen `target` used in other frameworks, but that's fine with me. Will update it...

src/pf-tooltip/pf-tooltip.component.js
priley86: +1 for making those functions private.

src/pf-tooltip/pf-tooltip.component.js
priley86: The problem I was facing with this is `appendChild` does not return the actual DOM element reference (and instead returns the fragment). The idea here was to get the actual DOM reference after it was appended, that way it could be styled subsequently and referenced as `this.tooltip` after. Querying the last `.tooltip` element added to the page seemed to be the easiest way, but if we change the container or order that may not be the case. While this works (and I've tested hovering all of the different tooltips really quickly in different orders), there may be a better approach (such as querying the last `.tooltip` added within the container). Thoughts? Happy to research this further ...

src/pf-tooltip/index.html
priley86: I'd rather avoid another plugin for this sort of thing if possible. The main reason I added this was to showcase the ability to re-render content after the tooltip content changes. This will be useful from the downstream framework perspective, as you might decide to render the content with a binding and then subsequently change that binding's value. Ideally, an Angular or React or Framework X shim handles this completely. I can write up an example of this next...

In React it might look like this:
```
<Tooltip placement="left" targetSelector="#btn-left"> {this.state.tooltipContent} <Tooltip>
```

where `Tooltip` is a React Component which renders a `<pf-tooltip>` internally and then fires this event when its bound value changes. Transclusion here in the React component is simply just rendering with {this.props.children} as the passed in `this.state.tooltipContent`. It will automatically fire a React lifecycle event, like `componentDidUpdate` when this content changes (when in turn could fire the web components `handleContentChanged`) .  

Hope this makes sense...it should allow for a seamless integration in the React ecosystem that behaves exactly like other React components (only the React shim is disguising this and rendering a web component for shared encapsulation of view logic).

I would think something in Angular directive would be very similar...

src/pf-utils/pf-utils.js
dlabrecq: I don't feel strongly about the "native" name, but thinking separate files is probably best.

src/pf-tooltip/index.html
dlabrecq: I was mostly wondering if this was a test case? I assume we can detect changes without requiring someone publishing an event.

priley86:
@dlabrecq i've made updates to this PR based on our previous notes. please give this a second look when you have a chance. 

Also, i have created another repo, "react-patternfly-shims" for housing the React shims to interface with WCs in this repo. I will write a sample React shim/component for `Tooltip` and reference this PR and the `pf-tooltip` web component (this should provide some more context on the `handleContentChanged` event). The idea is that these shims will aid in the adoption of WCs and make their consumption easier (and more natural) alongside our target frameworks.

If anyone has a chance to attempt this on the Angular side, it would be greatly appreciated :)

priley86:
A React Tooltip shim has been added here: 
https://github.com/patternfly-webcomponents/react-patternfly-shims/blob/master/src/Tooltip/Tooltip.js

and tested here:
https://github.com/patternfly-webcomponents/react-static-pf-webcomponents/commit/3409c0e24f5ef6a9e619b4c58fda545057948496

This should work quite well within the React ecosystem...

src/pf-tooltip/index.html
```diff
@@ -0,0 +1,51 @@
...
+
+    <br/>
+    <br/>
+    <script>
+        setTimeout(function(){
+          var pfTooltipRight = document.querySelector('[targetSelector="#btn-right"]');
+          pfTooltipRight.innerHTML = "Right content changed!";
+
+          var event = new CustomEvent('handleContentChanged',{});
+          pfTooltipRight.dispatchEvent(event);
```
dlabrecq: Looking good!

I suppose my only issue is that the app developer must generate/dispatch the 'handleContentChanged' event. Having to notify each web component every time changes are made could be quite painful for a large app?

Considering we don't need to do this when changing attributes; I'm thinking, the web component should determine when its own content has changed. I know we don't do this anywhere else, but this could be a good case for a base class or maybe a util function all web components must implement?

For example, we started doing this with pfTabs using a MutationObserver. We're currently observing childList, but believe we can use "subtree: true" for innerHtml?

priley86: This is a really important point (and one of the main things I was questioning on this and other similar components). I have talked with other devs about this I'd consider pretty knowledgeable in the WC space too and I suppose this is my preferred option right now (not saying we have to use it globally, but I think it works well here and is the most optimized pattern right now). MutationObservers can be quite slow if overused and events should be able to process outside framework changes more directly. Also, I think events are closer to the eventual pattern we'd be able to use with "slotchanged" once we implement Shadow DOM.  I've noticed this is a pretty common pattern in the WC space.
https://github.com/webcomponents/gold-standard/wiki/Content-Changes

My thinking is that we could write similar shim in Angular that I have done for React, and have the directive shim observe attribute changes and then fire the event to notify the WC. This way when scope variables/attributes change, we can abstract the actual event logic away in the shim, and the consumer doesn't need to know about it.

I  am hoping to write up an example of this today.

dlabrecq: I actually read that particular document earlier, but looking more at its base class example. The slotchange approach seems to only be used with Shadow DOM -- don't know if we should be coding for that.

The Angular shim could fire a change event, but a shim observing innerHTML changes won't perform any better than if the functionality were in a web component? (Keeping in mind that innerHtml is not an attribute.) And that still leaves the reference implementation with a manual solution.

If overusing MutationObserver is an issue, this feature could be disabled by default. Then, if an app developer wants to update content for a given web component, it could simply be enabled via an attribute.

https://hacks.mozilla.org/2012/05/dom-mutationobserver-reacting-to-dom-changes-without-killing-browser-performance/

dlabrecq: Alternatively, an app developer could simply call a function, similar to setAttribute. For example:

pfTooltipRight.setInnerHTML("Right content changed!");

Now the web component is aware it has changed without requiring the app developer to create and fire an event. In this case, the component doesn't need to listen for events, nor do we need a MutationObserver.

Again, my issue is that the app developer must generate/dispatch a 'handleContentChanged' event. I don't feel that having the web component listen for events is the right approach here -- the component itself should understand it has changed.

That said, I'll defer to @bleathem  -- I just want to ensure we nail down the correct approach.

priley86: Per our earlier chat, the innerHTML was only in place here for example sake. A framework would render this content via a binding (like I demonstrated with the React example). I suppose given the option of exposing a method such as "setInnerHtml()" to the shim or using Mutation Observers, I'd choose the former for the perf gain. However this makes things a bit trickier i.m.o. because you'd have to stringify the template after bindings get applied (it works though). The method I've proposed with events would just expose this to the WC within its DOM after the framework finishes rendering inside of its host element. There is no DOM observation in this approach.

priley86: i.e. Your directive ends up looking like this to the consumer <pf-tooltip-shim>Hello {{world}} </pf-tooltip-shim> . We will likely want to decide on a naming convention that doesn't conflict with the WC itself thus I've added "-shim" to the name but we could easily adopt another convention or change the WC name.

priley86: Per conversation today, i'll add a method to call the event (as opposed to requiring consumers to fire the event). This should be a more consumable API to those consuming the component directly (without a shim). This approach leveraged with the Tooltip should be reusable in other components.

priley86:
@dlabrecq @bleathem  i've updated the example page with a `setInnerHtml` method and additionally added `aria-describedby` attributes for a11y compliance.

priley86:
bump...ready for re-review

dlabrecq:
LGTM.

Didn't hear back from @bleathem, but believe this can be merged?

bleathem:
Great work @priley86, and thanks for the thorough reviews @dlabrecq

PR #17, opened 2017-01-08
Tabs: prevent default click event

priley86:
* noted that this event from a Tab anchor click was causing page changes in the Angular demo app i've started. We should be able to safely prevent the default action of this event.


priley86:
this should be ready for review now @dlabrecq @dtaylor113

PR #20, opened 2017-02-20
update PF version and README tweaks

priley86:
* update PF version & README tweaks

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
   },
   "dependencies": {
     "font-awesome": "^4.7.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "patternfly-webcomponents",
-  "version": "0.0.7",
+  "version": "0.0.8",
```
dlabrecq: If we want to use the release automation, we don't need to bump the version number here.

dlabrecq:
LGTM

PR #22, opened 2017-02-27
change WeakMap to Map

priley86:
This should resolve an issue in IE11. Still not sure about IE10.

Reference: https://github.com/webcomponents/webcomponentsjs/issues/495


bleathem:
:+1:  looks good to me

priley86:
thanks @dlabrecq & @bleathem .. much appreciated. Please do let me know next release if we can release this (we can update PF version too at the same time I am thinking).

dlabrecq:
If we update the PF version, please sync the patternfly-eng-release version as well.

PR #23, opened 2017-03-08
pf-dropdown

sahil143:
Introducing `<pf-dropdown>` new Web component based on Bootstrap Dropdown.

### DIFF
--- a/src/pf-dropdown/index.html
+++ b/src/pf-dropdown/index.html
@@ -18,60 +18,60 @@
 
     <!-- Dropdown Button -->
     <pf-dropdown class="dropdown" id="dropdown1">
-        <button class="btn btn-primary dropdown-toggle" type="button" id="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      <button class="btn btn-primary dropdown-toggle" type="button" id="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
         Dropdown
         <span class="caret"></span>
         </button>
-        <ul class="dropdown-menu">
-          <li class="dropdown-header">Header</li>
-          <li><a href="#">Item 1</a></li>
-          <li><a href="#">Item 2</a></li>
-          <li class="disabled"><a href="#">Item 3</a></li>
-          <li><a href="#">Item 4</a></li>
-          <li class="divider"></li>
-          <li><a href="#">Item 5</a></li>
-        </ul>
+      <ul class="dropdown-menu">
+        <li class="dropdown-header">Header</li>
+        <li><a href="#">Item 1</a></li>
+        <li><a href="#">Item 2</a></li>
+        <li class="disabled"><a href="#">Item 3</a></li>
+        <li><a href="#">Item 4</a></li>
+        <li class="divider"></li>
+        <li><a href="#">Item 5</a></li>
+      </ul>
     </pf-dropdown>
 
     <h2>Dropup</h2>
     <!-- Dropup Button -->
-    <pf-dropdown class="dropup" id="dropdown1">
-        <button class="btn btn-primary dropdown-toggle" type="button" id="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    <pf-dropdown class="dropup" id="dropdown2">
+      <button class="btn btn-primary dropdown-toggle" type="button" id="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
         Dropup
         <span class="caret"></span>
         </button>
-        <ul class="dropdown-menu">
-          <li class="dropdown-header">Header</li>
-          <li><a href="#">Item 1</a></li>
-          <li><a href="#">Item 2</a></li>
-          <li class="disabled"><a href="#">Item 3</a></li>
-          <li><a href="#">Item 4</a></li>
-          <li class="divider"></li>
-          <li><a href="#">Item 5</a></li>
-        </ul>
+      <ul class="dropdown-menu">
+        <li class="dropdown-header">Header</li>
+        <li><a href="#">Item 1</a></li>
+        <li><a href="#">Item 2</a></li>
+        <li class="disabled"><a href="#">Item 3</a></li>
+        <li><a href="#">Item 4</a></li>
+        <li class="divider"></li>
+        <li><a href="#">Item 5</a></li>
+      </ul>
     </pf-dropdown>
 
     <h2>Dropdown Kebab</h2>
     <!-- Dropdown Kebab -->
-    <pf-dropdown class="dropdown dropdown-kebab-pf" id="dropdown2">
-        <button class="btn btn-link dropdown-toggle" type="button" id="button2" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+    <pf-dropdown class="dropdown dropdown-kebab-pf" id="dropdown3">
+      <button class="btn btn-link dropdown-toggle" type="button" id="button2" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
         <span class="fa fa-ellipsis-v"></span>
         </button>
-        <ul class="dropdown-menu">
-          <li class="dropdown-header">Header</li>
-          <li><a href="#">Item 1</a></li>
-          <li><a href="#">Item 2</a></li>
-          <li class="disabled"><a href="#">Item 3</a></li>
-          <li><a href="#">Item 4</a></li>
-          <li class="divider"></li>
-          <li><a href="#">Item 5</a></li>
-        </ul>
+      <ul class="dropdown-menu">
+        <li class="dropdown-header">Header</li>
+        <li><a href="#">Item 1</a></li>
+        <li><a href="#">Item 2</a></li>
+        <li class="disabled"><a href="#">Item 3</a></li>
+        <li><a href="#">Item 4</a></li>
+        <li class="divider"></li>
+        <li><a href="#">Item 5</a></li>
+      </ul>
     </pf-dropdown>
 
     <h2>Dropup Kebab</h2>
     <!-- Dropup Kebab -->
-    <div class="col-xs-6 col-sm-4 col-md-2">    
-    <pf-dropdown class="dropup pull-right dropdown-kebab-pf" id="dropdown2">
+    <div class="col-xs-6 col-sm-4 col-md-2">
+      <pf-dropdown class="dropup pull-right dropdown-kebab-pf" id="dropdown4">
         <button class="btn btn-link dropdown-toggle" type="button" id="button2" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
         <span class="fa fa-ellipsis-v"></span>
         </button>
@@ -84,7 +84,7 @@
           <li class="divider"></li>
           <li><a href="#">Item 5</a></li>
         </ul>
-    </pf-dropdown>
+      </pf-dropdown>
 
     </div>
 
--- a/src/pf-dropdown/index.html
+++ b/src/pf-dropdown/index.html
@@ -1,0 +1,107 @@
+<!doctype html>
+<html>
+
+<head>
+  <title>pf-dropdown example</title>
+  <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/patternfly/3.13.0/css/patternfly.css">
+  <link rel="stylesheet" href="/dist/css/patternfly.css">
+  <link rel="stylesheet" href="/dist/css/patternfly-webcomponents.css">
+  <script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.22/webcomponents.js"></script>
+  <script src="/dist/js/patternfly.js"></script>
+</head>
+
+<body>
+  <div class="container">
+    <div class="page-header">
+      <h1>Dropdown</h1>
+    </div>
+
+    <!-- Dropdown Button -->
+    <pf-dropdown class="dropdown" id="dropdown1">
+        <button class="btn btn-primary dropdown-toggle" type="button" id="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+        Dropdown
+        <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu">
+          <li class="dropdown-header">Header</li>
+          <li><a href="#">Item 1</a></li>
+          <li><a href="#">Item 2</a></li>
+          <li class="disabled"><a href="#">Item 3</a></li>
+          <li><a href="#">Item 4</a></li>
+          <li class="divider"></li>
+          <li><a href="#">Item 5</a></li>
+        </ul>
+    </pf-dropdown>
+
+    <h2>Dropup</h2>
+    <!-- Dropup Button -->
+    <pf-dropdown class="dropup" id="dropdown1">
+        <button class="btn btn-primary dropdown-toggle" type="button" id="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+        Dropup
+        <span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu">
+          <li class="dropdown-header">Header</li>
+          <li><a href="#">Item 1</a></li>
+          <li><a href="#">Item 2</a></li>
+          <li class="disabled"><a href="#">Item 3</a></li>
+          <li><a href="#">Item 4</a></li>
+          <li class="divider"></li>
+          <li><a href="#">Item 5</a></li>
+        </ul>
+    </pf-dropdown>
+
+    <h2>Dropdown Kebab</h2>
+    <!-- Dropdown Kebab -->
+    <pf-dropdown class="dropdown dropdown-kebab-pf" id="dropdown2">
+        <button class="btn btn-link dropdown-toggle" type="button" id="button2" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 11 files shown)


priley86:
cc @bleathem @dlabrecq

bleathem:
This looks really good, thanks for the contribution @sahil143 !

src/pf-dropdown/pf-dropdown.component.js
```diff
@@ -0,0 +1,164 @@
...
+   *
+   * @param {Event} event
+   */
+  _keyHandler(event) {
+
+    event.preventDefault();
+    event.stopPropagation();
+
+    if (event.keyCode === 27 || event.keyCode === 13) {
+      this._clearDropdown();
```
dlabrecq: I can see how escape would clear, but why enter. Wouldn't we want enter to make a selection?

dlabrecq: In addition, we need to handle IE, Edge, etc. as well. For example, I grabbed this code from the table's filter which uses event.keyCode or event.which.

Sorry about the indentation.

ctx._pfFilter.filterInput.on("keypress", function (e) {
      var keycode = (e.keyCode ? e.keyCode : e.which);
      if (keycode === 13) {
	e.preventDefault();
        if (this.value.trim().length > 0) {
          addFilter(dt, {
            column: ctx._pfFilter.filterColumn,
            name: ctx._pfFilter.filterName,
            value: this.value
          });
        }
        return false;
      }
      return true;
    });

src/pf-dropdown/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "pf-dropdown",
+  "version": "1.0.0",
+  "description": "the pf-dropdown webcomponent",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
dlabrecq: We should start implementing unit tests with new components. Otherwise, this will be left for someone else to pick up.

src/pf-dropdown/pf-dropdown.component.js
```diff
@@ -0,0 +1,164 @@
...
+ */
+
+export class PfDropdown extends HTMLElement {
+
+  /**
+   * Called when an instance was inserted into the document
+   */
+  attachedCallback() {
+    let self = this;
+    this._button = this.querySelector('[data-toggle="dropdown"]');
```
sahil143: I have changed it to use class btn.

src/pf-dropdown/pf-dropdown.component.js
```diff
@@ -0,0 +1,164 @@
...
+
+    document.addEventListener('click', (event) => {
+      if (event.target !== this && !this.contains(event.target)) {
+        this._clearDropdown();
+      }
+    });
+
+    document.addEventListener('keydown', (event) => {
+      if (!/(38|40|27|13)/.test(event.keyCode) || /input|textarea/.test(event.target.tagName)) {
+        return;
```
sahil143: I have made some changes. Please take a look them. I hope it is more clear now.

src/pf-dropdown/pf-dropdown.component.js
```diff
@@ -0,0 +1,172 @@
...
+      }
+      if (this._disabled) {
+        return;
+      }
+      let keycode = event.keyCode ? event.keyCode : event.which;
+      let active = /\bopen/.test(this._button.parentNode.className);
+
+      //check if dropdown is open and keys other than 38,40,27,13 are not being used
+      if (active && /(38|40|27|13)/.test(keycode)) {
+        this._keyHandler(event, keycode);
```
dlabrecq: Do we need to test these keys here? I see they are tested in keyHandler as well.

src/pf-dropdown/pf-dropdown.component.js
priley86: +1 for some comments here on all these keycodes @sahil143 .

src/pf-dropdown/pf-dropdown.component.js
```diff
@@ -0,0 +1,172 @@
...
+  _keyHandler(event, keycode) {
+    event.preventDefault();
+    event.stopPropagation();
+
+    if (keycode === 27) {
+      this._clearDropdown();
+      this._button.focus();
+    }
+    if (keycode === 13) {
+      event.target.dispatchEvent(new MouseEvent('click'));
```
dlabrecq: Why is the original event canceled, only to dispatch a new event here?

Instead of cancelling all events up front, it may be best to move that code to the if statements which need it. Or, move it to  a function and call it as needed?

src/pf-dropdown/pf-dropdown.component.js
```diff
@@ -0,0 +1,172 @@
...
+      // index: guide focus on menu items
+      let index = Array.prototype.indexOf.call(menuItem, event.target);
+
+      if (keycode === 38 && index > 0) {
+        index--;
+      }
+      if (keycode === 40 && index < menuItem.length - 1) {
+        index++;
+      }
+      if (!~index) {
```
dlabrecq: It's not clear why we're inverting bits here? I'm guessing the index becomes negative at some point.

If I understand correctly, a ~9 value would become -10. If that's what's happening here, I'm certain tilda is what we want to test. Perhaps we can simply test if index < 0?

src/pf-dropdown/pf-dropdown.component.js
```diff
@@ -0,0 +1,172 @@
...
+    this._button.parentNode.classList.remove('open');
+    this.dispatchEvent(new CustomEvent('hidden.bs.dropdown', {}));
+  }
+
+  /**
+   * Support for phone browser
+   *
+   */
+  _detectTouch() {
+    if ('ontouchstart' in document.documentElement) {
```
dlabrecq: Curious how did you test this touch code?

I tried viewing the test page on my phone and the components don't render at all. I see the demo pages, but no web components. However, I can view https://patternfly-webcomponents.github.io/components.html.

I just want to ensure we don't introduce functionality that cannot be tested.

sahil143: I tried it on my phone using USB Debugging and port forwarding from chrome. Dropdown Web component seemed to be working fine That's why I added this functionality.

src/pf-dropdown/pf-dropdown.component.js
sahil143: When I used `preventDefault()` and `stopPropagation()` Default Enter events ceased to work that's why I had to define it manually. I didn't think about using `preventDefault()` only for navigation i.e. up/down keys. Maybe it will work in up/down `if` statement I'll give it a try and let you know.

src/pf-dropdown/pf-dropdown.component.js
sahil143: I used Tilda if at certain point index became negative. The only problem that arose until now was that initially when dropdown open and user press up key index value becomes undefined due to which nothing is focused on the menu. Yes, this problem can be easily solved by changing if condition to index < 0. Tilda is just a precaution.

src/pf-dropdown/pf-dropdown.component.js
sahil143: Not necessarily, I thought it would be good to check all keys before calling `keyHandler` method
if these keys are not pressed method will not be called. Nothing special.

priley86:
@sahil143 we will also need to add kebab support (this is really just different css). Let's discuss this next week 😄

src/pf-dropdown/pf-dropdown.component.js
sahil143: It should be working fine now. There was an error because at first, I was using `<pf-dropdown class="dropdown">` but to keep bootstrap API structure intact I changed it to `<pf-dropdown><div class="dropdown"> .... </div></pf-dropdown>`

src/pf-dropdown/pf-dropdown.component.js
sahil143: I have made some changes, Please take a look at them and let me know on further improvements.

src/pf-dropdown/index.html
```diff
@@ -0,0 +1,115 @@
...
+
+<body>
+  <div class="container">
+    <div class="page-header">
+      <h1>Dropdown</h1>
+    </div>
+
+    <!-- Dropdown Button -->
+    <pf-dropdown id="dropdown1">
+      <div class="dropdown">
```
priley86: This div should not be needed, use pf-dropdown element for that class instead (and ensure it styles the same as div) ;)

src/pf-dropdown/index.html
```diff
@@ -0,0 +1,115 @@
...
+          <li class="divider"></li>
+          <li><a href="#">Item 5</a></li>
+        </ul>
+      </div>
+    </pf-dropdown>
+
+    <h2>Dropup</h2>
+    <!-- Dropup Button -->
+    <pf-dropdown id="dropdown1">
+      <div class="dropup">
```
priley86: same here

src/pf-dropdown/index.html
```diff
@@ -0,0 +1,115 @@
...
+          <li><a href="#">Item 5</a></li>
+        </ul>
+      </div>
+    </pf-dropdown>
+
+    <h2>Dropdown Kebab</h2>
+    <!-- Dropdown Kebab -->
+    <pf-dropdown id="dropdown2">
+      <div class="dropdown dropdown-kebab-pf">
+        <button class="btn btn-link dropdown-toggle" type="button" id="button2" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
priley86: same here, no div needed, lets add class `dropdown-kebab-pf` to `pf-dropdown`

priley86:
This looks good to me. 👍  

@dlabrecq @bleathem anything else on this we'd like?

src/pf-dropdown/pf-dropdown.spec.js
```diff
@@ -0,0 +1,30 @@
...
+      //close dropdown
+      customElement.toggle();
+    }, false);
+
+    customElement.addEventListener('hidden.bs.dropdown', function () {
+      expect(button.parentNode.classList.contains('open')).toBe(false);
+      done();
+    }, false);
+  });
+});
```
dlabrecq: This is a good start, but it would be great to have a couple tests here. For example, to ensure that the user can make a selection. It should be easy enough to select one of the drop down anchors and test its expected text. You could even fire a click event on that selection.

Another test case could be to ensure the disabled option cannot be selected. That is, you shouldn't be able to fire a click event for any disabled menu options.

src/pf-dropdown/pf-dropdown.component.js
dlabrecq: I spoke with @priley86 about using the tilda. Please replace the tilda with something more like "index < 0" to ensure we're testing an expected value. It's also more readable for anyone maintaining the component.

src/pf-dropdown/pf-dropdown.spec.js
sahil143: Sure will add these tests and let you know when I'm done

src/pf-dropdown/pf-dropdown.component.js
priley86: yes, i think `index < 0` is more legible. This is minor, but I would agree here. +1

src/pf-dropdown/pf-dropdown.spec.js
```diff
@@ -24,7 +24,34 @@ describe("Patternfly Dropdown Component Test", function () {
...
+  it("disabled button should not open dropdown", function () {
+    var button = customElement.querySelector('.btn');
+    button.classList.add('disabled');
+    customElement.toggle();
+    expect(button.parentNode.classList.contains('open')).toBe(false);
+    button.dispatchEvent(new MouseEvent('click'));
+    expect(button.parentNode.classList.contains('open')).toBe(false);
+  });
+
+  it("should select an element", function () {
```
dlabrecq: Verifying a menu option is also a good test, but I was expecting the "selection" test to use a mouse click.

src/pf-dropdown/pf-dropdown.spec.js
```diff
@@ -24,7 +24,34 @@ describe("Patternfly Dropdown Component Test", function () {
...
+    var buttonText = customElement.querySelector('.btn');
+    var item = customElement.querySelector('ul.dropdown-menu > li:nth-child(2)');
+    expect(buttonText.innerText).toBe('Dropdown');
+    expect(item.innerText).toBe('Item 1');
+  });
+
+  it("should not click disabled item", function () {
```
dlabrecq: Instead of creating an event and dispatching yourself, I was expecting to use a mouse click. For example:
item.click();

You can add your own event listener to verify the selection.

src/pf-dropdown/pf-dropdown.spec.js
```diff
@@ -24,7 +24,34 @@ describe("Patternfly Dropdown Component Test", function () {
...
+    var item = customElement.querySelector('ul.dropdown-menu  li.disabled a');
+    var event = new MouseEvent('click',{
+      bubbles: true,
+      cancelable: true,
+      view: window
+    });
+    customElement.disableClick();
```
dlabrecq: Don't understand why you're disabling the element's click? Just evoke the click() yourself.

src/pf-dropdown/pf-dropdown.spec.js
sahil143: Would it be good to dispatch a customEvent in beforeEach function on clicked item and listen to that event in "selection" test to ensure the selection of item.

src/pf-dropdown/pf-dropdown.spec.js
sahil143: click() does not behave like a real mouse click on the anchor tag. I had to create a new mouse event(cancelable: true) to test the preventDefault() which prevents the URL from changing. And also it's easy to expect event.defaultPrevented on this new event. What do you think?

src/pf-dropdown/pf-dropdown.spec.js
sahil143: I had to use it because somehow in PhantomJS did not trigger `preventDefault()` action on disabled items. I had to call `disableClick()` method from `component.js` explicitly. I didn't have to call `disableClick()` while I was debugging in chrome.

dlabrecq: You shouldn't call disableClick() yourself. That wouldn't be a true test for a disabled menu option.

I don't believe this is an issue with PhantomJs. Other than the disabled class on the "li" element, I don't see anything that will prevent the anchor from generating an onclick event across browsers. You may also need to disable the anchor itself. For example:

&lt;li class="disabled">&lt;a href="#" onclick="return false;">Item 3&lt/a>&lt;/li>

src/pf-dropdown/pf-dropdown.spec.js
dlabrecq: Try testing an enabled menu option, first. You should be able to fire a click event and test for changes (e.g., which option was selected). You can even attach an event listener and test if it was called.

Creating your own event is not a true test of the underlying functionality. That is, testing event.defaultPrevented only indicates that you called preventDefault() yourself. 

I'm thinking the anchor is not disabled properly? See below.

sahil143:
@dlabrecq @priley86  I have updated the tests. Please take a look when you get a chance.
 The main problem with it was tests executed before the custom element initialized. I talked with @priley86 and fixed this. 
This time I used another approach to ensure the selection of enabled option and ensure that disabled options are not selected
 A custom event is fired when an enabled option is clicked and that custom event is not fired when disabled option is clicked.

src/pf-dropdown/pf-dropdown.spec.js
```diff
@@ -0,0 +1,72 @@
...
+
+  it("should not click disabled item", function () {
+    var item = customElement.querySelector('ul.dropdown-menu  li.disabled a');
+    var result = false;
+    // custom event is not fired for disabled item
+    customElement.addEventListener('itemClicked', function () {
+      result = true;
+    });
+    item.click();
+    expect(item.onclick()).toEqual(false);
```
dlabrecq: Good job on this test. Not certain we need to call/test onClick here?

src/pf-dropdown/pf-dropdown.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    }, false);
+    done();
+  });
+
+  it("disabled button should not open dropdown", function () {
+    var button = customElement.querySelector('.btn');
+    button.classList.add('disabled');
+    customElement.toggle();
+    expect(button.parentNode.classList.contains('open')).toBe(false);
+    button.dispatchEvent(new MouseEvent('click'));
```
dlabrecq: I'd still like to see this can be rewritten without dispatching an event yourself. Similar to the tests you've written below. We should be able to call click() on the button element.

sahil143: Yeah sure, It works same with `button.click()` .. Will push the changes in next commit.

src/pf-dropdown/pf-dropdown.spec.js
sahil143: Just to check that `item.onclick()` return  false for disabled items. Should I remove this line?

dlabrecq: Unless you're going to test this for enabled as well, it's probably unnecessary. We're already getting the callback.

sahil143:
@dlabrecq just pushed the changes you requested. Anything else we would like on this.

Issue #27, opened 2017-03-09
Convert v0 CEs to v1


priley86:
This will involve adopting the latest polyfill and converting all v0 CEs to v1 specification

Agathver:
I was reading through the [specs](https://www.w3.org/TR/custom-elements/) and [MDN](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Custom_Elements) and looks like it is not much work in adopting the source code, while there are not many components to change.

priley86:
@Agathver if you would like to try this out, please for go it! We will want to update the polyfill dependency and update the lifecycles (like you mentioned). I am hearing there may be a few bugs still in v1 (since it is so new), but we can surely test this out. Bleeding edge!

Let me know if I can help, but if you want to take a stab at this, feel free @Agathver !

Agathver:
> According to the spec, Custom Elements must be ES6 classes (https://html.spec.whatwg.org/multipage/scripting.html#custom-element-conformance).
https://github.com/webcomponents/webcomponentsjs

This is what they people say.

Custom Elements v1 are slowly landing across browsers and since we too are moving to Custom Elements v1 soon, we may need to rethink the packaging process that happens by webpack and babel, before switching to CEv1.

Polymer website mentions Safari Tech Preview and Chrome throw errors when CE v1 elements are used using ES5 Syntax.

PR #33, opened 2017-03-22
PatternFly 5 Modal

dabeng:
@bleathem @priley86 @dlabrecq  Would you like to review this PR again?

[Demo on dist branch](https://rawgit.com/dabeng/patternfly-webcomponents/modal-dist/src/pf-modal/index.html)

[Using pf-modal in ReactJS](https://plnkr.co/edit/ftdfVA?p=preview)

[Using pf-modal in Angular 1.6](https://plnkr.co/edit/JQu9BX?p=preview)

[Using pf-modal in Angular 2](https://plnkr.co/edit/WOLGXU?p=preview)

FYI, currently Firefox doesn't support attributeChange callback of Custom Elements v1. Please refer to https://bugzilla.mozilla.org/show_bug.cgi?id=1334051 for more details.

### DIFF
--- a/src/pf-modal/index.html
+++ b/src/pf-modal/index.html
@@ -1,0 +1,105 @@
+<!doctype html>
+<html>
+<head>
+    <title>pf-modal example</title>
+    <link rel="stylesheet" href="../../dist/css/patternfly.css">
+    <link rel="stylesheet" href="../../dist/css/patternfly-additions.css">
+    <link rel="stylesheet" href="../../dist/css/patternfly-webcomponents.css">
+</head>
+<body>
+<div class="container">
+    <div class="page-header">
+      <h1>Modals</h1>
+    </div>
+    
+    <button class="btn btn-default" id="btn-toggle-modal">Launch demo modal</button>
+    <pf-modal targetSelector="#btn-toggle-modal" backdrop keyboard>
+      <pf-modal-dialog>
+        <pf-modal-content>
+          <pf-modal-header modalTitle="Modal Title"></pf-modal-header>
+          <pf-modal-body>
+            <form class="form-horizontal">
+              <div class="form-group">
+                <label class="col-sm-3 control-label" for="textInput">Field One</label>
+                <div class="col-sm-9">
+                  <input type="text" id="textInput" class="form-control"/>
+                </div>
+              </div>
+              <div class="form-group">
+                <label class="col-sm-3 control-label" for="textInput2">Field Two</label>
+                <div class="col-sm-9">
+                  <input type="text" id="textInput2" class="form-control"/>
+                </div>
+              </div>
+              <div class="form-group">
+                <label class="col-sm-3 control-label" for="textInput3">Field Three</label>
+                <div class="col-sm-9">
+                  <input type="text" id="textInput3" class="form-control"/>
+                </div>
+              </div>
+            </form>
+          </pf-modal-body>
+          <pf-modal-footer cancelCaption="Cancel" saveCaption="Save"></pf-modal-footer>
+        </pf-modal-content>
+      </pf-modal-dialog>
+    </pf-modal>
+
+    <br/>
+    <br/>
+    <label>Events: </label>
+    <br/>
+    <textarea id="console" rows="15" cols="50" readonly style="resize: none;"></textarea>
+</div>
+<script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/1.0.0/webcomponents-lite.js"></script>
+<script src="//rawgit.com/webcomponents/custom-elements/master/src/native-shim.js"></script>
+<script src="../../dist/js/patternfly.js"></script>
+<script type="text/javascript">
+  (function() {
+    var modal = document.querySelector('pf-modal');
+    var console = document.querySelector('#console');
+    function monitorModalEvents (e) {
+      console.value += e.type + ' fired\n';
+    }
+    // It's a little trick for outputing unambiguous event trigger sequence
+    function getAdjustedSeq (msg) {
+      var logs = console.value.split('\n');
+      var seq = '';
+      if (logs[logs.length - 2].indexOf('hide fired') > -1) {
+        seq = console.value.split('\n').slice(0,-2).join('\n') +
+            '\n- - - ' + msg + ' - - -\npf-modal.hide fired\n';
+      } else {
+        seq = console.value + '- - - ' + msg + ' - - -\n';
+      }
+      return seq;
+    }
+    window.onload = function () {
+      document.querySelector('pf-modal-header .pf-hide-modal').addEventListener('click', function () {
+        console.value += '- - - cancel button of pf-modal-header is clicked - - -\n';
+      }, false);
+      document.querySelector('pf-modal-footer .pf-hide-modal').addEventListener('click', function () {
+        console.value += '- - - cancel button of pf-modal-footer is clicked - - -\n';
+      }, false);
+      modal.addEventListener('click', function (e) {
+        if (modal.backdrop && e.target === modal) {
+          console.value = getAdjustedSeq('Region outside of modal dialog is clicked');
+        }
+      }, false);
+      modal.addEventListener('keydown', function (e) {
+        if (!modal.keyboard && modal.open) {
+          return;
+        }
+        var isEscape = (e.key && e.key === 'Escape') || (e.keyIdentifier && e.keyIdentifier === 'U+001B') || (e.keyCode && e.keyCode === 27) || (e.which && e.which === 27);
+        if (isEscape) {
+          console.value = getAdjustedSeq('Escape key is pressed');
+        }
+      }, false);
+      modal.addEventListener('pf-modal.show', monitorModalEvents, false);
+      modal.addEventListener('pf-modal.shown', monitorModalEvents, false);
+      modal.addEventListener('pf-modal.hide', monitorModalEvents, false);
+      modal.addEventListener('pf-modal.hidden', monitorModalEvents, false);
+
+    };
+  }());
+</script>
+</body>
+</html>
--- a/src/pf-modal/pf-modal.component.js
+++ b/src/pf-modal/pf-modal.component.js
@@ -1,0 +1,293 @@
+import {pfUtil} from 'pf-utils.js';
+import PfModalDialog from 'pf-modal-dialog.component';
+import PfModalContent from 'pf-modal-content.component';
+import PfModalHeader from 'pf-modal-header.component';
+import PfModalBody from 'pf-modal-body.component';
+import PfModalFooter from 'pf-modal-footer.component';
+
+/*
+ * <b>&lt;pf-modal&gt;</b> element for Patternfly Web Components
+ *
+ * @example {@lang xml}
+ * <pf-modal targetSelector="#btn-toggle-modal" backdrop keyboard>
+ *   <pf-modal-dialog>
+ *     <pf-modal-content>
+ *       <pf-modal-header modalTitle="Modal Title"></pf-modal-header>
+ *       <pf-modal-body>custom content</pf-modal-body>
+ *       <pf-modal-footer></pf-modal-footer>
+ *     </pf-modal-content>
+ *   </pf-modal-dialog>
+ * </pf-modal>
+ *
+ * @prop {string} targetSelector Indicating which element will fireup the modal
+ * @prop {boolean} backdrop Indicating whether Clicking the backdrop could hide the modal or not
+ * @prop {boolean} keyboard Indicating whether clicking the escape key could hide the modal or not
+ * @prop {boolean} open Indicating whether or not the modal is opend
+ *
+ * @methods
+ *
+ * @show Show the modal
+ * @hide Hide the modal
+ * @toggle Toggle the visible/invisible state of modal
+ *
+ * @events
+ *
+ * @pf-modal.show It's fired immediately when the show instance method is called. If caused by a click,
+ * the clicked element is available as the relatedTarget property of the event.detail
+ * @pf-modal.shown It's fired when the modal has been made visible to the user (will wait for CSS transitions to complete). If caused by a click,
+ * the clicked element is available as the relatedTarget property of the event.detail
+ * @pf-modal.hide It's fired immediately when the hide instance method has been called
+ * @pf-modal.hidden It's fired when the modal has finished being hidden from the user (will wait for CSS transitions to complete)
+ */
+
+export class PfModal extends HTMLElement {
+
+  /*
+   * Get the show/hide state of modal
+   *
+   * @returns {boolean} True if modal is visible
+   */
+  get open () {
+    return this.hasAttribute('open');
+  }
+
+  /*
+   * Set flag indicating modal is shown
+   *
+   * @param {boolean} val True to set modal visible
+   */
+  set open (val) {
+    if (this.open !== val) {
+      if (val) {
+        this.setAttribute('open', '');
+      } else {
+        this._hideModal();
+      }
+    }
+  }
+
+  /*
+   * Get the backdrop setting
+   *
+   * @returns {boolean} True if users are allowed to hide modal on clicking backdrop
+   */
+  get backdrop () {
+    return this.hasAttribute('backdrop');
+  }
+
+  /*
+   * Set flag indicating clicking backdrop of modal could hide modal
+   *
+   * @param {boolean} val True to enable backdrop clicking
+   */
+  set backdrop (val) {
+    if (this.backdrop !== val) {
+      if (val) {
+        this.setAttribute('backdrop', '');
+      } else {
+        this.removeAttribute('backdrop');
+      }
+    }
+  }
+
+  /*
+   * Get the keyboard setting
+   *
+   * @returns {boolean} True if users are allowed to hide modal when escape key is pressed
+   */
+  get keyboard () {
+    return this.hasAttribute('keyboard');
+  }
+
+  /*
+   * Set flag indicating pressing escape key could hide modal
+   *
+   * @param {boolean} val True to enable escape key reaction
+   */
+  set keyboard (val) {
+    if (this.keyboard !== val) {
+      if (val) {
+        this.setAttribute('keyboard', '');
+      } else {
+        this.removeAttribute('keyboard');
+      }
+    }
+  }
+
+  /*
+   * Only attributes listed in the observedAttributes property will receive this callback
+   */
+  static get observedAttributes() {
+    return ['targetselector', 'open', 'keyboard', 'backdrop'];
+  }
+
+  /*
+   * Called when element's attribute value has changed
+   *
+   * @param {string} attrName The attribute name that has changed
+   * @param {string} oldValue The old attribute value
+   * @param {string} newValue The new attribute value
+   */
+  attributeChangedCallback (attrName, oldValue, newValue) {
+    if (attrName === 'targetselector' && newValue !== null && oldValue === null) {
+      this._target = document.querySelector(newValue);
+      if (this._target) {
+        this._target.addEventListener('click', this.show.bind(this), false);
+        this._target.setAttribute('data-bound', 'bound');
+      }
+    }
+    if (attrName === 'open' && newValue !== null) {
+      this._showModal();
+    }
+    if (attrName === 'keyboard') {
+      if (newValue === '') {
+        this.addEventListener('keydown', this._keydownHandler, false);
+      } else {
+        this.removeEventListener('keydown', this._keydownHandler, false);
+      }
+    }
+  }
+
+  /*
+   * An instance of the element is created or upgraded
+   */
+  constructor () {
+    super();
+
+    this._mask = null;
+  }
+
+  /*
+   * Called every time the element is inserted into the DOM
+   */
+  connectedCallback () {
+    this.className = 'modal fade';
+    this.setAttribute('tabindex', -1);
+
+    if (this.open) {
+      this._showModal();
+    }
+
+    if (this.getAttribute('targetSelector')) {
+      this._target = document.querySelector(this.getAttribute('targetSelector'));
+      if (this._target && !(this._target.getAttribute('data-bound') === 'bound')) {
+        this._target.addEventListener('click', this.show.bind(this), false);
+      }
+    }
+
+    this.addEventListener('click', this.hide, false);
+  }
+
+  /*
+   * Show the modal when attribute open is added or property open is enabled
+   * @private
+   */
+  _showModal () {
+    pfUtil.addClass(document.body, 'modal-open');
+    this._mask = document.createElement('div');
+    this._mask.className = 'modal-backdrop fade';
+    document.body.appendChild(this._mask);
+    pfUtil.reflow(this._mask);
+    pfUtil.addClass(this._mask, 'in');
+    pfUtil.reflow(this);
+    pfUtil.addClass(this, 'in');
+    pfUtil.once(this.querySelector('pf-modal-dialog'), 'transitionend', this._afterShowModal, this);
+  }
+
+  /*
+   * Show the modal
+   * @public
+   */
+  show (e) {
+    if (e) {
+      this.dispatchEvent(new CustomEvent('pf-modal.show', { 'detail': { 'relatedTarget': e.currentTarget }}));
+      this._triggeredByUser = true;
+    } else {
+      this.dispatchEvent(new CustomEvent('pf-modal.show', {}));
+    }
+    if (this.open) {
+      return;
+    }
+    this.open = true;
+  }
+
+  /*
+   * Callback after modal is shown
+   * @private
+   */
+  _afterShowModal () {
+    this.focus();
+    if (this._triggeredByUser) {
+      this.dispatchEvent(new CustomEvent('pf-modal.shown', { 'detail': { 'relatedTarget': this._target }}));
+    } else {
+      this.dispatchEvent(new CustomEvent('pf-modal.shown', {}));
+    }
+    this._triggeredByUser = false;
+  }
+
+  /*
+   * Hide the modal when attribute open is removed or property open is disabled
+   * @private
+   */
+  _hideModal () {
+    pfUtil.removeClass(document.body, 'modal-open');
+    pfUtil.removeClass(this._mask, 'in');
+    pfUtil.removeClass(this, 'in');
+    pfUtil.once(this, 'transitionend', this._afterHideModal, this);
+  }
+
+  /*
+   * Hide the modal
+   * @public
+   */
+  hide (e) {
+    if (e) {
+      e.preventDefault();
+      if (!(pfUtil.getClosest(e.target, '.pf-hide-modal') || (e.target === e.currentTarget && this.backdrop))) {
+        return;
+      }
+    }
+
+    this.dispatchEvent(new CustomEvent('pf-modal.hide', {}));
+
+    if (!this.open) {
+      return;
+    }
+    this._hideModal();
+  }
+
+  /*
+   * Callback after the modal is hidden
+   * @private
+   */
+  _afterHideModal () {
+    this.removeAttribute('open');
+    if (this._mask) {
+      this._mask.remove();
+      this._mask = null;
+    }
+    this.dispatchEvent(new CustomEvent('pf-modal.hidden', {}));
+  }
+
+  /*
+   * Toggle the visible/invisible state of modal
+   * @public
+   */
+  toggle () {
+    return this.open ? this.hide() : this.show();
+  }
+
+  /*
+   * Handler of keydown event of escape key
+   * @private
+   */
+  _keydownHandler (e) {
+    let isEscape = (e.key && e.key === 'Escape') || (e.keyIdentifier && e.keyIdentifier === 'U+001B') || (e.keyCode && e.keyCode === 27) || (e.which && e.which === 27);
+    if (isEscape && this.open) {
+      this.hide();
+    }
+  }
+
+}
+
+window.customElements.define('pf-modal', PfModal);
--- a/src/pf-modal/pf-modal.spec.js
+++ b/src/pf-modal/pf-modal.spec.js
@@ -1,0 +1,158 @@
+describe ('PatternFly Moadl Component Tests', function () {
+  var customElement, button;
+
+  function render(html) {
+    var template = document.createElement('template');
+    template.innerHTML = html;
+    document.body.appendChild(document.importNode(template.content, true));
+  }
+
+  beforeEach(function (done) {
+    var mo = new MutationObserver(function () {
+      mo.disconnect();
+      done();
+    });
+    mo.observe(document.body, { childList: true });
+    render(
+      '<button id="btn">Launch demo modal</button>' +
+      '<pf-modal id="modal" targetSelector="#btn" backdrop keyboard>' +
+      '  <pf-modal-dialog>' +
+      '    <pf-modal-content>' +
+      '      <pf-modal-header></pf-modal-header>' +
+      '      <pf-modal-body>Modal body</pf-modal-body>' +
+      '      <pf-modal-footer></pf-modal-footer>' +
+      '    </pf-modal-content>' +
+      '  </pf-modal-dialog>' +
+      '</pf-modal>'
+    );
+    button = document.querySelector('#btn');
+    customElement = document.querySelector('#modal');
+  });
+
+  afterEach(function () {
+    document.body.removeChild(button);
+    button = null;
+    document.body.removeChild(customElement);
+    customElement = null;
+    if (document.querySelector('.modal-backdrop')) {
+      document.body.removeChild(document.querySelector('.modal-backdrop'));
+    }
+  });
+
+  describe('Prperties & Attributes testing', function () {
+    it('Reflecting properties to attributes', function () {
+      customElement.open = true;
+      expect(customElement.getAttribute('open')).toBe('');
+
+      customElement.backdrop = false;
+      expect(customElement.getAttribute('backdrop')).toBe(null);
+
+      customElement.backdrop = true;
+      expect(customElement.getAttribute('backdrop')).toBe('');
+
+      customElement.keyboard = false;
+      expect(customElement.getAttribute('keyboard')).toBe(null);
+
+      customElement.keyboard = true;
+      expect(customElement.getAttribute('keyboard')).toBe('');
+    });
+
+    describe('Set open to false', function () {
+      it('_hideModal() should be called when open is disable', function () {
+        spyOn(customElement, '_hideModal');
+        customElement.setAttribute('open', '');
+        customElement.open = false;
+        expect(customElement._hideModal).toHaveBeenCalled();
+      });
+    });
+
+    it('Reflecting attributes to properties', function () {
+      customElement.setAttribute('open', '');
+      expect(customElement.open).toBe(true);
+
+      customElement.removeAttribute('open');
+      expect(customElement.open).toBe(false);
+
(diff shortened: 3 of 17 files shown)


src/pf-modal/pf-modal.component.js
```diff
@@ -0,0 +1,254 @@
...
+        this._element.removeEventListener('keydown', this._keydownHandler, false);
+      }
+    }
+  }
+
+  /*
+   * An instance of the element is created or upgraded
+   * We build up the complete DOM tree of web components
+   */
+  constructor () {
```
priley86: We aren't quite here yet with v1 constructors. Can you use the v0 attachedCallback for now? Would like to change all these new components in a new PR for #27

priley86:
wow! this is a great first start @dabeng. 

Can you do me a favor and go ahead and expand the iframe size in `app.html` for this PR?  Ideally we can change it to something like:

```
    function resizeIframe(obj) {
      obj.style.height = obj.contentWindow.document.body.scrollHeight + 200 + 'px';
```

This is a really promising start. We will definitely need some tests (please see my last merged PR for notes on how to debug these locally with Karma/Phantom). 

Also, maybe a modal with a form like the one [here](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/modals.html) would be nice?


You may want to ask @dgutride and @dlabrecq for any advice here too.

src/pf-modal/pf-modal.component.js
dabeng: got it. no question

dabeng:
You mentioned " _Also, maybe a modal with a form like the one here would be nice?_"

There are three modal trigger buttons that will popup three different types of modal. The "backdrop" feature of No.2 modal is disabled. The "keyboard" feature of No.3 modal is disabled.

dabeng:
Hi @priley86 , I have added unit testing and replaced custom elements v1 with v0. Please review agagin 😊

priley86:
wow! this is awesome @dabeng !!!  😈 

Thanks for making the v0 change. Looks good!

I have one more additional request after reviewing a bit more. Can we pretty please with sugar on top change the `<div modalPart="header">` to `<pf-modal-header class="modal-header">`? 

I think it's fine to keep the Bootstrap template separate like you have done, we would just need to query this new `pf-modal-header` custom element in the `createdCallback` of `pf-modal` instead and then copy its content into your `div class="modal-header` in the rendered template. We could also allow the user to override the class here too w/ the `pf-modal-header` CE (i.e. we could default its rendered class to `modal-header` if nothing is provided, but if a class is provided from the user, we could override with that). 

I would suggest the same patterns for `pf-modal-body` and `pf-modal-footer`.

I know this is a little bit of refactor, but take your time. You're doing an awesome job with this already!! 👍

priley86:
also - you may want to check your tests in build w/ Phantom too? Travis seems to be upset again.

src/pf-modal/pf-modal.component.js
```diff
@@ -0,0 +1,250 @@
...
+    pfUtil.addClass(this._element, 'in');
+    pfUtil.once(this._dialog, 'transitionend', this._afterShowModal, this);
+  }
+
+  /*
+   * Show the modal
+   * @public
+   */
+  show () {
+    this._element.dispatchEvent(new CustomEvent('modalShow', { 'detail': { 'relatedTarget': this._target }}));
```
dabeng: What about "show.pf.modal"?

dabeng: After all, bs is bootstrap keyword.

dabeng:
After running "gulp test", all unit tests succeed. I don't know why travis failed.

Agathver:
@dabeng, Travis will fail if any of these tasks fails: `lint`, `test`, `doc`
Try running all the three gulp tasks.
And there also could be runtime errors in Travis like, out of memory.

src/pf-modal/pf-modal.component.js
priley86: You raise a great point. We have been using the same events as Bootstrap before (so we'd have to adopt this convention globally), but I'm not opposed to this.

I will raise a question on this to the mailing list this week. Feel free to pledge support for this...

karma.conf.js
```diff
@@ -20,7 +20,9 @@ module.exports = function(config) {
       'node_modules/jasmine-promises/dist/jasmine-promises.js',
       'node_modules/webcomponentsjs/full.js',
       'dist/js/patternfly.js',
-      'src/**/*.spec.js'
+      'src/**/*.spec.js',
+      'dist/css/patternfly.css',
+      'dist/css/patternfly-webcomponents.css'
```
Agathver: Please do not include CSS files here. PhantomJS goes out of memory when you try to load CSS and WebFonts in Travis.

dabeng:
Hi @priley86 , I thought about custom element pf-modal-header, but I found that it violated the aims of Custom Element -- reusability. Except for being included in pf-modal, are there other scenarios where pf-modal-header can apply in?

In fact, what we need is just a container which contains custom content for header(body or footer). We don't even care what container element. Div, Section, even a are all right. It's really a fuss to make a standalone custom element just for a wrapper element.

src/pf-modal/pf-modal.component.js
```diff
@@ -0,0 +1,250 @@
...
+  _showModal () {
+    pfUtil.addClass(document.body, 'modal-open');
+    this._mask = document.createElement('div');
+    this._mask.className = 'modal-backdrop fade';
+    document.body.appendChild(this._mask);
+    pfUtil.reflow(this._mask);
+    pfUtil.addClass(this._mask, 'in');
+    pfUtil.reflow(this._element);
+    pfUtil.addClass(this._element, 'in');
+    pfUtil.once(this._dialog, 'transitionend', this._afterShowModal, this);
```
Agathver: You will need to manually call `afterShowModal` in your tests.

karma.conf.js
dabeng: Where should I place these two css file references?

src/pf-modal/pf-modal.component.js
dabeng: I did so already

karma.conf.js
Agathver: Do not include them at all. You probably need CSS for transitions, instead of depending upon the events, manually call the handlers

src/pf-modal/pf-modal.component.js
Agathver: Then removing your CSS files will probably make the travis errors go away.

To exactly simulate Travis, make an ubuntu 14.04 lxc container with 1.5gb memory allocated to it and run the tests :smiling_imp:

src/pf-modal/pf-modal.component.js
dabeng: Hi @priley86 , "show.bs.modal" is the special event namesapce of jquery style. Are you sure to introduce it into patternfly web components?

dabeng: got it

priley86:
@dabeng I think the main idea with `pf-modal-header`, etc. is to make the DOM more expressive. There is a really good article on this [here](https://www.pluralsight.com/blog/software-development/html5-web-components-overview). 

For too long we have encouraged "div" soup and this project aims to avoid that. Ideally, we can still support the "div" scenarios by using common attributes or properties as selectors, but what I am suggesting is that `divs` are no longer necessary in many cases. Really they should only be needed when the consumer is extending the template with a LOT of custom style (typically they would just needed to change a class on the `pf-modal-header` though). But if we use a common selector in `pf-modal` though, the CE `pf-modal-header` along with a `div` are both valid, however one is obviously more expressive.

Does this sort of make more sense why I am suggesting this now? I see your point about reusability of child CEs, however I much prefer the "sugar" of having a more expressive syntax (and typically these children CEs are pretty simple to implement). Any more thoughts here?

src/pf-modal/pf-modal.component.js
```diff
@@ -0,0 +1,250 @@
...
+    this.open = true;
+  }
+
+  /*
+   * Callback after modal is shown
+   * @private
+   */
+  _afterShowModal () {
+    this._element.focus();
+    this._element.dispatchEvent(new CustomEvent('shown.bs.modal', { 'detail': { 'relatedTarget': this._target }}));
```
dlabrecq: We should document exactly what events, names, and properties are generated. If not for this private function, at least for the public ones. And perhaps also at the top of the component's JsDoc to make it clear what public events are available?

src/pf-modal/pf-modal.component.js
```diff
@@ -0,0 +1,250 @@
...
+    pfUtil.addClass(this._element, 'in');
+    pfUtil.once(this._dialog, 'transitionend', this._afterShowModal, this);
+  }
+
+  /*
+   * Show the modal
+   * @public
+   */
+  show () {
+    this._element.dispatchEvent(new CustomEvent('show.bs.modal', { 'detail': { 'relatedTarget': this._target }}));
```
priley86: +1 for standardizing these. Thus the question 😉

src/pf-modal/pf-modal.component.js
```diff
@@ -0,0 +1,250 @@
...
+  toggle () {
+    return this.open ? this.hide() : this.show();
+  }
+
+  /*
+   * Handler of keydown event of escape key
+   * @private
+   */
+  _keydownHandler (e) {
+    let isEscape = (e.key && e.key === 'Escape') || (e.keyIdentifier && e.keyIdentifier === 'U+001B') || (e.keyCode && e.keyCode === 27);
```
dabeng: I have tested this PR on IE10 or Edge. It worked well.

Besides, there are too many code snippets of patternfly-webcomponents waiting for cross-browser checkout. keydown event of "Escape" key is just the tip of the iceberg. That's why I wanna introduce jquery as a common module at the beginning.

dabeng:
Hi @priley86 , I have add children CEs(pf-modal-header, pf-modal-body and pf-modal-footer)

src/pf-modal/index.html
```diff
@@ -0,0 +1,57 @@
...
+</head>
+<body>
+<div class="container">
+    <div class="page-header">
+      <h1>Modals</h1>
+    </div>
+    <hr>
+    
+    <button class="btn btn-default" id="btn-toggle-modal">Launch demo modal</button>
+    <pf-modal targetSelector="#btn-toggle-modal" backdrop keyboard>
```
dabeng: I think "provide all this raw HTML themselves" is just what users want most from pf-modal because children custom elements(pf-modal-header, pf-modal-body, pf-modal-footer) provide users with the maximal free space to construct their own internal structure of pf-modal 😊

src/pf-modal/pf-modal.component.js
dabeng: jQuery is just a Module providing us with the syntactic sugar of basic DOM operation. It's essentially different with angular, react, vuejs. It's not essentially different with vanilla javascript if we just make use of its core DOM functionalities of jquery.

dabeng: There is a reason that Bootstrap 4 doesn't give up jQuery

src/pf-modal/index.html
dlabrecq: I disagree. The Angular Patternfly modal was created because developers don't want to have to find an example and provide the raw HTML themselves. Otherwise, there is little value in creating a component.

src/pf-modal/pf-modal.component.js
dlabrecq: There are cases where event.keyCode does not exist.

dlabrecq: Many Angular developers won't use the components because of jQuery -- it's a non-starter for some people.

priley86: Hi @dabeng, i think it is important to keep in mind that our core two target frameworks initially will be the Angular JS and React JS frameworks. These are still the most common frameworks in use at Red Hat today. What @dlabrecq is saying is certainly true about many of these communities avoiding the use of Jquery. It adds additional API overhead (where frameworks often come in solving the same problems) and typically is not as clean as fit as framework native components or now, web components. You might have seen this link on our PF Showcase page about how to avoid jquery:
http://codeblog.cz/vanilla/overview.html#prerequisities

With the amount of existing components implementing complex behaviors with Jquery (such as our treeview), we obviously have a long ways to go to remove it and reimplement these components. But we've learned a lot from some of the communites below on how to do this... 

http://www.vanillalist.com/
http://thednp.github.io/bootstrap.native/

src/pf-modal/index.html
priley86: could we not provide different variations of the `pf-modal` CE, such as an `pf-about-modal` which inherits or reimplements this CE? @dlabrecq @dabeng

priley86: i too was thinking about this with something such as a "USA State dropdown" or a "USA Zip code select" component? These types of common UI components may become a possibility with web components...

priley86:
thanks @dabeng , i think this looks good, but will wait for others to weigh in on this. 

Have we decided on an event naming convention @bleathem ? do we prefer `show.pf.hidden`,etc. ? Maybe we can get another PR going to clean this up (I'm partially responsible for this inconsistency!) ;0

src/pf-modal/index.html
dabeng: @priley86 @dlabrecq @bleathem The core functionality of pf-modal is providing a togglable dialog. It's just a dialog wrapper with show/hide effect and doesn't care what's internal structure of dialog. Can we come to an agreement about this point ?

dabeng: I think the value of web components should not include completing stuff outside its duty. e.g. custom content or layout.

dabeng: <pf-modal>
  <pf-modal-header>custom content</pf-modal-header>
  <pf-modal-body>custom content</pf-modal-body>
  <pf-modal-footer>custom content</pf-modal-footer>
<pf-modal>

dabeng: let modal = new PfModal({
  headerInfo: {…},
  bodyInfo: {…},
  footerInfo: {…},
  otherOptions: {…}
});
document.body.appendChild(modal);

dabeng: There gotta be a better one.

src/pf-modal/pf-modal.component.js
dabeng: Hi @dlabrecq , I have added ``` event.which``` testing.

dabeng:
@priley86 @dlabrecq @bleathem , there is one other point I would like to discuss with you.

Some web components are suitable for declaration with html, like pf-tabs. Some are suitable for instantiation with javascript. And others can be created by any of two above ways.

Here, I think it's a better way to create pf-modal with declaration pattern because the way can facilitate users to construct the internal content of pf-modal's every sections.

dlabrecq:
There was a discussion during the last UXD/UI Developer meeting and believe the consensus is that we want to provide a pattern's default behavior first, but allow developers to override if necessary.

For example, the pf-modal-header component can provide the close button and title layout by default. If a developer wishes to override that, they can replace it using transclusion via the body of pf-modal-header. 

Likewise, the pf-modal-footer component should provide the save and cancel buttons by default and then allow developers to override if necessary. The pf-modal-body component is fine as is.

dabeng:
Cool. This consensus is a great improvement.

dabeng:
I have added "override" feature for pf-modal-header and pf-modal-footer, hope current commit can archive your requirements @priley86 @dlabrecq @bleathem

src/pf-modal/index.html
```diff
@@ -0,0 +1,48 @@
+<!doctype html>
+<html>
+<head>
+    <title>pf-tooltip example</title>
+    <link rel="stylesheet" href="/dist/css/patternfly.css">
```
priley86: you'll want to make all of these `dist/css/` paths relative after the recent update. i.e. `../../dist/css/patternfly.css`. Check the other components for an example on this...

src/pf-modal/index.html
```diff
@@ -0,0 +1,48 @@
+<!doctype html>
+<html>
+<head>
+    <title>pf-tooltip example</title>
```
priley86: should be: `pf-modal` example

priley86:
nice improvement @dabeng, works for me. Can you do me a huge favor and make paths relative, and then test your test page with Travis? I left instructions in #38 for how to do this... please just post your test page link here once done...

src/pf-modal/pf-modal.spec.js
```diff
@@ -0,0 +1,139 @@
...
+      expect(callback).toHaveBeenCalled();
+    });
+
+    it('hidden.bs.modal should be triggered', function () {
+      customElement.addEventListener('hidden.bs.modal', callback);
+      customElement._afterHideModal();
+      expect(callback).toHaveBeenCalled();
+    });
+  });
+
```
dlabrecq: Now that we're providing default cancel, save, and close buttons, we should test the events here. For example, we can test if clicking the close button hides the modal. Users should be able to receive an event for save, cancel, etc.

src/pf-modal/index.html
dabeng: Sorry for my careless

src/pf-modal/pf-modal.component.js
```diff
@@ -0,0 +1,275 @@
...
+ * @toggle Toggle the visible/invisible state of modal
+ *
+ * @events
+ *
+ * @show.bs.modal It's fired immediately when the show instance method is called. If caused by a click,
+ * the clicked element is available as the relatedTarget property of the event.detail
+ * @shown.bs.modal It's fired when the modal has been made visible to the user (will wait for CSS transitions to complete). If caused by a click,
+ * the clicked element is available as the relatedTarget property of the event.detail
+ * @hide.bs.modal It's fired immediately when the hide instance method has been called
+ * @hidden.bs.modal It's fired when the modal has finished being hidden from the user (will wait for CSS transitions to complete)
```
dabeng: got it

src/pf-modal/index.html
```diff
@@ -0,0 +1,48 @@
...
+    <link rel="stylesheet" href="/dist/css/patternfly.css">
+    <link rel="stylesheet" href="/dist/css/patternfly-additions.css">
+    <link rel="stylesheet" href="/dist/css/patternfly-webcomponents.css">
+    <script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.22/webcomponents.js"></script>
+    <script src="/dist/js/patternfly.js"></script>
+</head>
+<body>
+<div class="container">
+    <div class="page-header">
+      <h1>Modals</h1>
```
dabeng: Sorry @dlabrecq , I can't understand this requirement. Could you explain it in more details?

src/pf-modal/index.html
dabeng: **Test page of pf-modal** -- https://rawgit.com/dabeng/patternfly-webcomponents/modal-dist/src/pf-modal/index.html

dabeng:
@priley86 , Test page of pf-modal -- https://rawgit.com/dabeng/patternfly-webcomponents/modal-dist/src/pf-modal/index.html

src/pf-modal/index.html
```diff
@@ -0,0 +1,48 @@
...
+          </div>
+          <div class="form-group">
+            <label class="col-sm-3 control-label" for="textInput3">Field Three</label>
+            <div class="col-sm-9">
+              <input type="text" id="textInput3" class="form-control">
+            </div>
+          </div>
+        </form>
+      </pf-modal-body>
+      <pf-modal-footer></pf-modal-footer>
```
dabeng: @dlabrecq , I can't find your mentioned textarea on http://www.patternfly.org/angular-patternfly/#/api/patternfly.modals.directive:pfAboutModal

dlabrecq: I wasn't specifically speaking about the modal example, but something more like listView, toolbar, etc. Click on the buttons in the example below and you'll see the interaction.

http://www.patternfly.org/angular-patternfly/#/api/patternfly.views.directive:pfListView

src/pf-modal/index.html
dlabrecq: Please disregard that comment. I thought I was looking at the pf-modal-header. It's difficult to read when GitHub injects all these comments between lines of code.

src/pf-modal/index.html
dabeng: OK. No question.

priley86:
Thanks @dabeng. The modal looks to work, but it seems that your branch needs rebasing? If I go her for instance: 
https://rawgit.com/dabeng/patternfly-webcomponents/modal-dist/index.html

It does not load the same files as it does in `master`. It should look like this one...with your modal added:
https://rawgit.com/patternfly-webcomponents/patternfly-webcomponents/master-dist/index.html

dabeng:
I'll check this issue ASAP after I append enough unit testing.

dabeng:
@priley86 , Is pf.modal.show final custom event namespace convention?

dlabrecq:
I thought the events were going to be <component name>.<event name> (e.g., pf-modal.show)?

dlabrecq:
I thought the events were going to be &lt;component name&gt;.&lt;event name&gt; (e.g., pf-modal.show)?

priley86:
I typically see events namespaced with a period but I'm fine with a hyphen as it lines up closer with our component names. @dlabreque

dabeng:
@priley86 https://rawgit.com/patternfly-webcomponents/patternfly-webcomponents/master-dist/index.html can display well.

dabeng:
@dlabrecq , I have appended monitor console to demonstrate how to capture custom events dispatched from pf-modal.

dlabrecq:
@dabeng I like the monitor console you added.

dlabrecq:
@priley86 I recommended using the component name as the event prefix because of component names like pf-list-view, pf-utilization-bar-chart, etc. I don't feel strongly about the dot syntax, but seems odd using pf.utilization.bar.chart.show.

priley86:
@dlabrecq same - i don't have a strong preference (either convention should work). Let's go with hyphens 👍

src/pf-modal/index.html
```diff
@@ -0,0 +1,95 @@
...
+    <link rel="stylesheet" href="../../dist/css/patternfly-webcomponents.css">
+    <script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.22/webcomponents.js"></script>
+    <script src="../../dist/js/patternfly.js"></script>
+</head>
+<body>
+<div class="container">
+    <div class="page-header">
+      <h1>Modals</h1>
+    </div>
+    <hr>
```
priley86: @dabeng  Can we remove this `<hr>` ? It's superfluous after the `page-header` div (and shows borders instead of one). Otherwise...i think this looks great. Nice work!

dabeng: done

dabeng:
@priley86 @dlabrecq Custom events of pf-modal have been changed according to <component name>.<event name> (e.g., pf-modal.show)

priley86:
hey @dabeng, one other thing I'd like to test w/ this is handling content changes (most commonly a framework will render nested bindings inside of your CE's template). What's important to note  about this is framework lifecycle exists outside of CE lifecycle, so one way simple way to handle this is to expose a `handleContentChanged` event in each CE which could have content changes via the framework. There is an example of this in the `pf-tooltip`. You can test this w/ a framework, but the easiest way is to just query the DOM and manipulate it via JS, and then trigger the event (also in the pf-tooltip test page). Can we do something like this here too?

dabeng:
@priley86  I understand your intentions. You wanna dynamically change the content of CE after instantiation or declaration. 

However, from the perspective of functionality flexibility, I think setInnerHtml() method can not do better than reInitialize() method and reInitialize() method cannot do better than new CE().

Specifically, setInnerHtml() method doesn't take into account attributes' changes of CE and reInitialize() would be a chicken ribs compared with new CE().

All told, I think both setInnerHtml() method and handleContentChanged event are not necessary.

dabeng:
@dlabrecq @bleathem, please share your opinions about setInnerHtml() or handContentChanged event.

dlabrecq:
I vaguely recall having this discussing with @priley86, @bleathem, and other devs a while back. I believe the issue was uncovered during a code review of pf-tooltip and related to how the (Angular) component wrapper supported transcluded content. Component wrappers push properties down to the component, but must respond to component changes by listening for events.

In the particular case of pf-tooltip, we also added the setInnerHtml function to apply the new, transcluded content. This gave us the ability to encapsulate the handleContentChanged event, which is needed by the component wrappers.

That said, we don't have many components right now and even fewer that support transclusion. However, we should continue to follow the patterns we've been implementing across all components.

priley86:
Great job w/ this @dabeng ...appreciate your patience and hope these patterns make sense. It is actually quite a bit of work to write well tested CE's which mimic that of other open source framework components, but the payoff will come later when we can consume them globally and change their behavior in one core place ;)

dabeng:
@priley86 @dlabrecq , thanks a lot for your consecutive improvement advises. 

Now I still have a question. If users don't make use of setInnerHtml() to modify the content of pf-tooltip, then handleContentChanged event will not be triggered.

In fact, there are so many ways to change the content of CE. It's endless to manually dispatch custom events.

Maybe we should introduce **DOMSubtreeModified** into CE definition. Personally, I think it's a good replacer for handleContentChanged event.

dlabrecq:
The best we can do is document this as the recommended way to set transcluded content for Patternfly components. If users choose not to follow the docs, we cannot stop them. 

The problem with DOMSubtreeModified is that it comes with a strong warning regarding its usage.

"Be very careful with this event it is easy to cause an infinite loop if you decide to change the DOM inside the event handler."

Agathver:
>The best we can do is document this as the recommended way to set transcluded content for Patternfly components. If users choose not to follow the docs, we cannot stop them.

Or use mutation observers? 

>So maybe we should introduce DOMSubtreeModified into CE definition. Personally, I think it's a good replacer for handleContentChanged event. In other words, handleContentChanged event should be derived from DOMSubtreeModified

@dabeng The entire DOM events have been [deprecated](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Mutation_events) and it comes with [performance issues](https://github.com/angular/material/issues/5681). And is a lot crashy, --  if there is an unhandled error inside the DOMSubtreeModified listener, the entire page might die!

dlabrecq:
Mutation observers have their own performance issues, which is why we're avoiding them here.

dabeng:
@priley86 @dlabrecq @Agathver setInnerHtml() method and matched "pf-modal.change" event are all added. Of course, related unit testings are supplemented too.

dabeng:
@Agathver , changing the DOM inside the DOMSubtreeModified handler remind me of calling setState() in render() under React.js.

priley86:
@dabeng wow! great job...the tests are very helpful and it looks like you've captured all of the events. 

I have one last request. Can we include the 'x' (close icon) in the top right of the demo page modal?
After this, I'm ready to approve...

priley86:
@dlabrecq any more thoughts on these updated unit tests?

priley86:
One other thing here... after looking at `setInnerHtml()` further, you are accepting html (which I think is fine), but it doesn't look like you are re-initializing the component? I am just noting that slight difference w/ the `pf-tooltip` which calls `init()` after `handleContentChanged` fires in the `attachedCallback`.

I'm not sure if it matters or not with your CE though...just want to be sure we can support the case of something like this (rendered in a framework):
```
<pf-modal>
<pf-modal-header>header</pf-modal-header>
<pf-modal-body> <p> Greetings earthlings, we come in {frameworkModel.salutation} </p> </pf-modal-body>
<pf-modal-footer>footer</pf-modal-footer>
</pf-modal>
```
Just want to be sure that when the pf-modal-body content changes via the framework binding (which occurs outside the CE's lifecycle). If it's not updating the content, I'd suggest allowing the framework to fire `handleContentChanged` to notify the CE to reinitialize.

Can we just verify this? You could also just write some html/js to test it like:
```
<pf-modal id='pf-modal'>
<pf-modal-header>header</pf-modal-header>
<pf-modal-body> <p> Greetings earthlings, we come in <span id='saluation'>peace</span> </p> </pf-modal-body>
<pf-modal-footer>footer</pf-modal-footer>
</pf-modal>
```
js:
```
var pfModal = document.getElementById('pf-modal');
document. getElementById('#salutation').textContent = 'search of web components';
pfModal.dispatchEvent(new CustomEvent('handleContentChanged', {}));
```

Does the content of the `pf-modal-body` then change?

dabeng:
@priley86 , the following code snippets can change the content of pf-modal-body. No `problem.
```js
var pfModal = document.getElementById('pf-modal');
document. getElementById('#salutation').textContent = 'search of web components';
```

Additionally, according to @dlabrecq's recommandation:

> The best we can do is document this as the recommended way to set transcluded content for Patternfly components. If users choose not to follow the docs, we cannot stop them.

In pf-modal, we just allow uses to fire handleContentChanged event by calling setInnerHtml() method.

priley86:
@dabeng do you mind creating a plunker demonstrating this (like I have done with `pf-tooltip`)? I think the next step is to test this out with Angular / React bindings...

dabeng:
@priley86 @dlabrecq hope you like the [Demo](https://codepen.io/dabeng/pen/LypNwr)
screen shot 2017-04-18 at 11 47 20 am

dabeng:
@priley86 
This is ** [Using pf-modal in ReacJS](http://codepen.io/dabeng/pen/QvyPbp)**
And this is [**Using pf-modal in Angular**](http://codepen.io/dabeng/pen/pPEjKz?editors=1010)

priley86:
thanks very much @dabeng for providing this... these templates are great and we can certainly use them in the future!

For React, I am really hoping to support something like this:
http://codepen.io/priley86/pen/qmaxwd

This gives an example of how content can change via props/bindings. This will definitely happen in real app scenarios where you are passing dynamic content via props/state into your modal (and it will have to update). I guess my hope is that we support this kind of behavior via Mutation Observer or providing some event to tell the WC to reload itself...

Thanks again for researching, this is great!!!

priley86:
in looking at this again, @dabeng, another option may be to "shim" the modal-header, modal-body, and modal-footer and just write the content as `this.props.children`. If this works, we should still get bindings, but you will have to test that it works with the WC... My guess is it is the same challenge, but I'm not sure...

dabeng:
Sorry @priley86 , I couldn't get your points, can you explain in more detail?

> write the content as this.props.children

dabeng:
As for pf-modal demo under anguarl 2 context. I think it's a good timing when we replace custom elements v0 with v1.

priley86:
Sorry @dabeng, it is hard to give a full explanation without a complete example... I think the simple use case is still valid (and a good first step to test). Can you make this WC support the codepen I forked? 
http://codepen.io/priley86/pen/qmaxwd

The issue is we need to make the WC respond to changes in state/props being passed down into the template of the modal. When I click "Change field one", I'm updating state (and props coming down), but the content label does not change. The content in the modal body will be assumed dynamic by the consumer (angular or react). 

The other idea I'm having is to create React shims (which internally render `pf-modal-mody`) and inside it's body we render `this.props.children`. That content is then dynamic based on what is transcluded by the consumer....

Long story short, if you can fix the use case for the codepen I've added (and reload the content either via Mutation Observer or an event), then that use case should work as well...(we'd just replicate the solution)...

Hope this helps...

dabeng:
Hi @priley86 , I have to use dangerousSetInnerHTML() method in order to render HTML string as real HTML in JSX. Maybe there are other better solutions. Please review again [Demo](http://codepen.io/dabeng/pen/QvyPbp?editors=1010)
As for "create React shims", @dlabrecq has denied the use of mutation 

> observer:Mutation observers have their own performance issues, which is why we're avoiding them here.

So I moved to use setInnerHTML() method and dispatch pf-modal.changed event.

priley86:
@dabeng have you seen @Agathver's latest plnkr? 
https://plnkr.co/edit/cGrMMm?p=preview

My vote is to do something similar here. This is part of the issue with setting nested content as a string. That is not the typical scenario by any React / Angular consumer I don't think. I.e. most React components I work with wrap nested components as `this.props.children` (and render them inside). Would like to stay consistent w/ this...

Can you take a look at this one and see what you think? We are still having an open conversation on this, so appreciate any feedback...

dabeng:
Hi @priley86 , you mean I should replace setInnerHTML() with mutation observer?

dabeng:
@priley86 For the last few days, I tried introduced MutationObserver object into the pf-modal. However, I found that it's not a good idea to build up web components based on mutation observer mechanism because it will result in a huge of redundant and low-performance code snippets as wile as poor readability and maintainability. Specially, it's a nightmare for complicated and nested web component to work with mutations.

Agathver:
@dabeng The entire Polymer project is heavily based on MutationObservers, and they have been stabilized for quite a time, so it can be safe to assume the performance issues of MutationObservers have been alleviated by now.

http://stackoverflow.com/questions/31659567/performance-of-mutationobserver-to-detect-nodes-in-entire-dom

Sure there is some performance hits (~3%), but the convenience it gives, is worth it, and in many cases it won't be the performance bottleneck either.

priley86:
@dabeng @Agathver my vote is to try to abstract the Observers away as best we can for now (can we possibly write a helper to contain that logic?). Either way, same performance issue is there for Angular...and I'm not sure of another way to fire the `handleContentChanged` event once Angular has finished rendering a binding. 

Long term, I think we look to `slotchanged` offered by the Shadow DOM slots.
https://github.com/webcomponents/gold-standard/wiki/Content-Changes

dabeng:
@priley86 I have made the pf-modal respond to changes in state/props being passed down into the template of the modal. Could you the demos on the top of this page again?

priley86:
@dabeng great job... Really like the examples across Angular/React you did! These are really useful for future components and show how flexible web components can be... 

PF-modal is definitely a more complex component which we can all learn from. I think the examples adequately show conditional rendering and you did a great job w/ this. The API is a bit confusing to me from the examples, so I definitely think writing a really doc page for this will be useful (clearly denoting different WC attributes/intended usages).

My previous plnkr appears to be updating correctly now, but styling is off and looks like there is some event handler error in console. Any idea? It doesn't appear we have a MutationObserver now, but the `pf-modal-body` seems to be rendered separately (works). So it may not be needed...
http://codepen.io/priley86/pen/qmaxwd

@cdcabrera @dlabrecq @bleathem any more thoughts before we move forward here? I think next step is to clearly document the API...would <3 a PR [here](https://github.com/patternfly-webcomponents/patternfly-webcomponents.github.io) documenting the PF Modal next...

dabeng:
@priley86 I added assistant elements(pf-modal-dialog and pf-modal-content), please see this demo
http://codepen.io/dabeng/pen/GmyxWy

priley86:
awesome @dabeng ! the new codepen looks great. thanks so much...

Let me know once you have some API docs ready for showcase site or if you want any help ;) I will definitely review those w/ you.

dabeng:
@priley86 ，I have upgraded pf-modal to v1, please review Demos again.

priley86:
looks good to me @dabeng ... the v1 examples appear to be working correctly in everything other than Firefox for me (as you noted). These also appear to be different issues than we ever had w/ v0. Is there an upstream issue we can track?  You're definitely the farthest along w/ v1 research so thanks a lot for pushing this forward...

Any short term workarounds for this? 

The only other thing I noted was a minor difference in the Cancel/Save buttons and our PF Core test page. The button heights are different for some reason. Can you double check that we don't need to extend this modal with any new styles?

https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/modals.html

priley86:
@dabeng my guess is `attributeChangedCallback` is currently in development for FF...

http://caniuse.com/#feat=custom-elementsv1

cdcabrera:
@dlabrecq @priley86 @dabeng merged this in, open issues accordingly

PR #37, opened 2017-04-04
Add editorconfig for consistent style across IDEs

Agathver:
[Editorconfig](http://editorconfig.org) allows for consistent code style across IDEs. It is supported by most of the IDEs out of the box (WebStorm and Atom fall into this category) or by using an official plugin (VS Code and Sublime).

This configuration standardizes the most basic configuration about whitespace, line endings and trailing spaces. 

This also renames `eslint.json` to `.eslintrc.json` (default eslint config file name), so as to allow integration with various IDE linting plugins


priley86:
I like this suggestion @Agathver. thanks for submitting!

bleathem:
this is great, thanks @Agathver!

PR #40, opened 2017-05-12
upgrade web components of patternfly to custom elements v1

dabeng:
@bleathem @priley86 , would you like to review this PR?

FYI, https://rawgit.com/dabeng/patternfly-webcomponents/ce-v1-dist/index.html

### DIFF
--- a/app/app.html
+++ b/app/app.html
@@ -5,7 +5,8 @@
   <link rel="stylesheet" href="../dist/css/patternfly.css">
   <link rel="stylesheet" href="../dist/css/patternfly-additions.css">
   <link rel="stylesheet" href="../dist/css/patternfly-webcomponents.css">
-  <script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.22/webcomponents.js"></script>
+  <script src="https://rawgit.com/webcomponents/webcomponentsjs/master/webcomponents-lite.js"></script>
+  <script src="https://rawgit.com/webcomponents/custom-elements/master/src/native-shim.js"></script>
   <script src="../dist/js/patternfly.js"></script>
   <script type="text/javascript">
     function resizeIframe(obj) {
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ env:
 before_install:
   - if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
   - npm install git+https://github.com/patternfly/patternfly-eng-release.git
+  - export CHROME_BIN=chromium-browser
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
 
 install: true
 
--- a/TESTS.md
+++ b/TESTS.md
@@ -1,37 +1,49 @@
 # Testing Patternfly-Webcomponents
 
 ### Unit Tests
-Unit tests are run with [Karma](https://karma-runner.github.io/1.0/index.html) and [karma-phantomjs-launcher](https://github.com/karma-runner/karma-phantomjs-launcher).
+Unit tests are run with [Karma](https://karma-runner.github.io/1.0/index.html) and [karma-chrome-launcher](https://github.com/karma-runner/karma-chrome-launcher).
 
 To run unit tests, simply run:
 ```
 gulp test
 ```
 
-To debug unit tests locally with karma test server and Chrome, first edit `karma.conf.js` to target Chrome:
+To debug unit tests locally with karma test server and Headless Chrome, first add the following setting snippets to `karma.conf.js`:
 ```
-  browsers: ['Chrome'],
+  browsers: ['HeadlessChrome'],
+
+  customLaunchers: {
+    HeadlessChrome: {
+      base: 'Chrome',
+      flags: ['--no-sandbox', '--headless', '--disable-gpu', ' --remote-debugging-port=9222']
+    }
+  }
 ```
+
+**Note**: --headless flag is only available Chrome 59+.
+
 You can also change `logLevel` for futher debug logging:
 ```
 logLevel: config.LOG_DEBUG,
 ```
 
- you can then run the following to start a Karma debug server:
- ```
- karma start karma.conf.js
- ```
-(then hit the "DEBUG" button in the top right to debug)
-
- To debug unit tests locally with Phantom JS, target Phantom:
- ```
-karma start --browsers PhantomJS_custom
+you can then run the following to start a Karma debug server:
+```
+karma start karma.conf.js
 ```
 
-This will launch a local PhantomJS debug server, which you can access here:
+then you can access `http://localhost:9876/` and hit the "DEBUG" button in the top right to debug
+
+In addtion, if your target browers include Firefox, Opera, Safari , IE and so on. You can append these names to the "browsers" option.
 ```
-http://localhost:9000/webkit/inspector/inspector.html?page=2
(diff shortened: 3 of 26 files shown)


app/app.html
```diff
@@ -5,7 +5,8 @@
   <link rel="stylesheet" href="../dist/css/patternfly.css">
   <link rel="stylesheet" href="../dist/css/patternfly-additions.css">
   <link rel="stylesheet" href="../dist/css/patternfly-webcomponents.css">
-  <script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/0.7.22/webcomponents.js"></script>
+  <script src="https://rawgit.com/webcomponents/webcomponentsjs/master/webcomponents-lite.js"></script>
```
Agathver: You may want to use webcomponents-loader.js in place of webcomponents-lite. It will more reflect the real world usage of web components

Agathver:
@dabeng I see you are using the bundled webcomponents-lite.js that comes with pf. It ships the v0 polyfill and thus is inoperable with v1 webcomponents. Use this instead: https://github.com/WebComponents/webcomponentsjs

priley86:
@dabeng great job getting us started w/ v1! It looks like you've updated the WCs to use the new v1 API. Can you please provide your Travis fork / test pages once done? I think we will definitely need to test the v1 components across browsers and see if anything has changed.

bleathem:
Nice work @dabeng.  Thanks for the review and suggestion @Agathver!

@priley86, @dlabrecq do we have any suggestions for @dabeng on how to get the tests to run?

dabeng:
@Agathver @priley86 Thanks a lot for your review. Could you give advice about how to get the tests to run?

Agathver:
@dabeng the polyfill you are including supports only the web components  v0 spec. Use v1 of the webcomponent polyfill via the webcomponents loader I mentioned above. This should make your tests pass if everything else is okay.

You need to download a copy of webcomponents -lite.js v1 from https://github.com/WebComponents/webcomponentsjs and reference it inside karma.conf.js

priley86:
Yep... agreed. I think we can safely move to the `@webcomponents/webcomponentsjs` package now and remove `webcomponentsjs` from package.json. That npm package is no longer current.

If you want to quickly test via CDN, it looks like those scripts exist here:
https://cdnjs.com/libraries/webcomponentsjs

dabeng:
@Agathver @priley86 , thank you for your great recommendations, but I still can't get rid of the following errors after I run "gulp test".

screen shot 2017-05-15 at 9 59 13 am

Agathver:
See: https://github.com/ariya/phantomjs/issues/13895
Apparently PhantomJS (a specific version of JavaScript core breaks if you define a property called "name" somewhere with `{ configurable: true }`) It's probably fixed on newer JSCore, but PhantomJS hasn't been updated in a while.

priley86:
@dabeng i think we should try to reference the polyfill from `node_modules` if possible. 
https://github.com/priley86/patternfly-webcomponents/commit/db4e931a70f7a92c8476f387a2a3bb873dfa2e98

I tried testing this branch in Karma w/ Chrome & Phantom JS just now... 

A couple observations:
1. Chrome seems to work fine (as usual) w/ v1 upgrades you've added. Safari seems to work as well (all components). For whatever reason, Tabs seem to break on Firefox now (properly something silly).
2. Phantom JS does indeed break when loading the v1 polyfill (I'm getting the same errors in debugger in the webcomponents-lite.js file provided). We should probably note this error and open an issue against the webcomponents repo & Phantom upstream repos...
screen shot 2017-05-16 at 11 29 58 am

FWIW, we've had so many problems w/ Phantom in the past that I'd be completely in favor of removing this restriction and leaving Phantom. It regularly does not properly represent the other browsers, and that's why it seems a lot of open source projects have moved to tools like [Sauce Labs](https://saucelabs.com/) which gives us Selenium...and most importantly, real browsers to test with in an automated way!

It's becoming a barrier to us moving forward w/ the spec...so I'm in favor of removing Phantom. We can of course keep tabs on this issue and revisit it though. We can also try to target `xvfb` like @Agathver was suggesting, or look into a new tool such as SauceLabs.

thoughts @bleathem ?

priley86:
further detail on the polyfill breakage...

It's using `Set`, and es2015 type. 
`var ma=new Set("annotation-xml `...
https://github.com/webcomponents/webcomponentsjs/blob/master/webcomponents-hi-ce.js

bleathem:
Can we look at using headless chrome instead of PhantomJS?
https://developers.google.com/web/updates/2017/04/headless-chrome
https://chromium.googlesource.com/chromium/src/+/lkgr/headless/README.md

Agathver:
Headless chrome is a great solution. Glad that Chromium is packaging it out-of-the box. But, we have to wait till Chrome 59.

Currently the pf dependencies also bring SlimerJS - which is used for headless testing in Firefox, may be we should be testing in Firefox too.

dabeng:
You are right. In fact, I recommend every developer of patternfly-webcomponents to use the following configuration snippets in karma.conf.js file:

> browsers: ['HeadlessChrome', 'FireFox', 'Opera', 'Safari', 'IE'],

At the same time, we need to install the following launchers firstly:

- Firefox (launcher requires karma-firefox-launcher plugin)
- Opera (launcher requires karma-opera-launcher plugin)
- IE (launcher requires karma-ie-launcher plugin)
- Safari (launcher requires karma-safari-launcher plugin)

Apparently, installing various browsers can not be imperative, so I didn't do that in this patch.

dabeng:
@priley86 , travis fork isn't generated this time due to the following error:

> Authentication failed for 'https://[secure]@github.com/dabeng/patternfly-webcomponents.git/'

Agathver:
Due to some recent security issue regarding accidental leaking of OAuth tokens in Travis build logs, GitHub has revoked all OAuth tokens used in Travis. You need to add a new personal access token in the  AUTH_TOKEN environment variable.

The process is highlighted here: https://github.com/patternfly-webcomponents/patternfly-webcomponents/blob/master/TESTS.md#enabling-travis-builds

bleathem:
@dlabrecq can you update the OAUTH token here too?

priley86:
this is great! let's merge...

priley86:
@dabeng one more minor request before i forget...can we update TESTS.md to reflect new changes for debugging Karma (as you noted in comments)?

dabeng:
@priley86 , please review TESTS.md :)

priley86:
thanks for the updates @dabeng ! great job... i think we still need to confirm w/ @dlabrecq if this can be merged.

dlabrecq:
The AUTH_TOKEN has been updated

dlabrecq:
The AUTH_TOKEN has been updated for patternfly-webcomponents. However, your personal AUTH_TOKEN still needs to be updated in order to run Travis and generate a master-dist branch for personal forks.

dabeng:
@priley86, please review https://rawgit.com/dabeng/patternfly-webcomponents/ce-v1-dist/index.html

priley86:
looks great @dabeng ! 

tested these in Safari, Chrome, and FF on my local, all worked fine. Also nice job w/ the `pf-tabs` fix. I have a downstream addition I'd like to add w/ support for secondary tabs once this is merged.

@dabeng @bleathem @dlabrecq  can we merge this now?

bleathem:
Nice work, thanks @dabeng!

PR #44, opened 2017-05-25
adds secondary tabs to pf-tabs

priley86:
minor tweak to `pf-tabs` to support secondary tabs. 

i've added a test page here just now:
https://rawgit.com/priley86/patternfly-webcomponents/secondary-tabs-dist/app/app.html?dir=pf-tabs&file=index.html

cc @bleathem @dabeng @sahil143

### DIFF
--- a/src/pf-tabs/index.html
+++ b/src/pf-tabs/index.html
@@ -1,22 +1,47 @@
 <!doctype html>
 <html>
+
 <head>
   <title>pf-tabs example</title>
   <link rel="stylesheet" href="../../dist/css/patternfly.css">
   <link rel="stylesheet" href="../../dist/css/patternfly-additions.css">
   <link rel="stylesheet" href="../../dist/css/patternfly-webcomponents.css">
 </head>
+
 <body>
   <div class="container">
     <div class="page-header">
       <h1>Tabs</h1>
     </div>
     <pf-tabs>
-      <pf-tab tabTitle="Tab1" active="true">
-        <p>Tab1 content here</p>
+      <pf-tab tabTitle="Tab One" active="true">
+        <p>Tab one content here</p>
       </pf-tab>
-      <pf-tab tabTitle="Tab2">
-        <p>Tab2 content here</p>
+      <pf-tab tabTitle="Tab Two">
+        <p>Tab two content here</p>
+      </pf-tab>
+    </pf-tabs>
+
+    <pf-tabs>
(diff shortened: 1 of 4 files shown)


priley86:
I will go ahead and shame myself publicly for using `!important` in the CSS. I'm not sure if there is any easy way to override this, or if we should add another contribution to Core...

src/pf-tabs/pf-tabs.scss
```diff
@@ -0,0 +1,23 @@
+pf-tabs.nav.nav-tabs .nav.nav-tabs-pf {
+  & > li {
+    font-size: 12px;
+
+    &:first-child > a {
+      padding-left: 15px;
+
+      &:before {
+        left: 15px!important;
```
srambach: It looks like if Patternfly removes the !important from the rule `.nav-tabs-pf>li:first-child>a:before` then this one isn't needed either.

src/pf-tabs/index.html
```diff
@@ -12,13 +12,36 @@
...
+          </pf-tab>
+          <pf-tab tabTitle="Secondary Tab Two">
+            <p>Secondary tab two content.</p>
+          </pf-tab>
+        </pf-tabs>
+      </pf-tab>
+      <pf-tab tabTitle="Tab Two">
+        <pf-tabs class="nav nav-tabs nav-tabs-pf">
+          <pf-tab tabTitle="Secondary Tab One" active="true">
+            <p>Secondary tab one content.</p>
```
cdcabrera: If the demo could have different nested tab labels and content under the different main tabs that might help highlight functionality. Right now it appears slightly confusing, as if the tabs aren't functioning properly

src/pf-tabs/index.html
```diff
@@ -12,13 +12,36 @@
...
+          </pf-tab>
+          <pf-tab tabTitle="Secondary Tab Two">
+            <p>Secondary tab two content.</p>
```
cdcabrera: Same as first comment...

priley86:
tab content updated. :)

@srambach it looks like we may be holding off on the PF core change? 

Anything else before this can be merged? I think we are safe...

src/pf-tabs/pf-tabs.scss
srambach: Until Patternfly removes the !important some day, this one will have to stay.

srambach:
Nothing else; !important will have to stay for the time being.

PR #49, opened 2017-06-28
Build Update, Unit Test & NPM Scripts

cdcabrera:
Updated to allow both Travis and users who don't have the necessary version of Chrome installed to run unit tests. Minor NPM script updates.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -60,7 +60,7 @@ module.exports = function(config) {
 
     // start these browsers
     // available browser launchers: https://npmjs.org/browse/keyword/karma-launcher
-    browsers: ['HeadlessChrome'],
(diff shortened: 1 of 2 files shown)


cdcabrera:
@Agathver kind of agree on letting everyone catch up... and definitely agree on adding other browser types for the unit testing. 

However the millisecond speed gains for 22+ local run unit tests shouldn't outweigh the possibility we may be hampering future contributions because we require unit tests that also happen to require a specific major version increment of Chrome to function properly. 

Going the route of forced, or expected, adoption isn't going to gain anything. Travis still has the benefits of the speed gains with this fix, except now we're also encouraging an ease in development and contribution that helps everyone, current and future.

PR #50, opened 2017-07-12
pf-popover

sahil143:
Introducing `pf-popover` based on Bootstrap Popover

### DIFF
--- a/src/pf-popover/index.html
+++ b/src/pf-popover/index.html
@@ -1,0 +1,105 @@
+<!doctype html>
+<html>
+
+<head>
+    <title>pf-popover example</title>
+    <link rel="stylesheet" href="../../dist/css/patternfly.css">
+    <link rel="stylesheet" href="../../dist/css/patternfly-additions.css">
+    <link rel="stylesheet" href="../../dist/css/patternfly-webcomponents.css">
+    <style>
+        #div {
+            margin-top: 40px;
+        }
+    </style>
+</head>
+
+<body>
+    <div class="container">
+        <div class="page-header">
+            <h1>popovers</h1>
+        </div>
+        <hr>
+
+            <pf-popover id="popover-right" placement="right" target-selector="#btn-right">
+                popover text on right popover text on right popover text on right popover text on right popover text on right
+            </pf-popover>
+            <button id="btn-right" type="button" class="btn btn-primary" aria-describedby="popover-right">use of public handler(toggle()) to open</button>
+
+        <div id="div">
+            <pf-popover id="popover-right" placement="right" target-selector="#btn-right-t-c" popover-title="Popver on right" dismissible="true">
+                popover text on right popover text on right popover text on right popover text on right popover text on right
+            </pf-popover>
+            <button id="btn-right-t-c" type="button" class="btn btn-primary" aria-describedby="popover-right">Popover on right with title and close button</button>
+        </div>
+
+        <div id="div">
+            <pf-popover id="popover-right" placement="right" target-selector="#btn-right-t" popover-title="Popver on right">
+                popover text on right popover text on right popover text on right popover text on right popover text on right
+            </pf-popover>
+            <button id="btn-right-t" type="button" class="btn btn-primary" aria-describedby="popover-right">popover on right with title</button>
+        </div>
+
+        <div id="div">
+            <pf-popover id="popover-right" placement="right" target-selector="#btn-right-c" dismissible="true">
+                popover text on right popover text on right popover text on right popover text on right popover text on right
+            </pf-popover>
+            <button id="btn-right-c" type="button" class="btn btn-primary" aria-describedby="popover-right">popover on right with close button</button>
+        </div>
+
+        <div id="div">
+            <pf-popover id="popover-right" placement="right" target-selector="#btn-right-n">
+                popover text on right popover text on right popover text on right popover text on right popover text on right
+            </pf-popover>
+            <button id="btn-right-n" type="button" class="btn btn-primary" aria-describedby="popover-right">Popover on right</button>
+        </div>
+
+        <hr>
+
+        <div id="div" style="margin-left: 300px;">
+            <pf-popover id="popover-left" placement="left" target-selector="#btn-left-t-c" popover-title="Popover on left" dismissible="true">
+                popover text on left popover text on left popover text on left popover text on left popover text on left
+            </pf-popover>
+            <button id="btn-left-t-c" type="button" class="btn btn-primary" aria-describedby="popover-left">Popover on left with title and close button</button>
+        </div>
+
+        <hr>
+
+        <div id="div">
+            <pf-popover id="popover-left" placement="left" target-selector="#btn-left-v" popover-title="Popover on left" dismissible="true">
+                popover text on left popover text on left popover text on left popover text on left popover text on left
+            </pf-popover>
+            <button id="btn-left-v" type="button" class="btn btn-primary" aria-describedby="popover-left">Popover on left but not in view</button>
+        </div>
+
+        <hr>
+
+        <div id="div">
+            <pf-popover id="popover-top" placement="top" target-selector="#btn-top-t-c" popover-title="Popover on top" dismissible="true">
+                popover text on top popover text on top popover text on top popover text on top popover text on top
+            </pf-popover>
+            <button id="btn-top-t-c" type="button" class="btn btn-primary" aria-describedby="popover-top">Popover on top with title and close button</button>
+        </div>
+
+        <hr>
+
+        <div id="div" style="margin-bottom: 100px">
+            <pf-popover id="popover-bottom" placement="bottom" target-selector="#btn-bottom-t-c" popover-title="Popover on bottom" dismissible="true">
+                popover text on bottom popover text on bottom popover text on bottom popover text on bottom popover text on bottom
+            </pf-popover>
+            <button id="btn-bottom-t-c" type="button" class="btn btn-primary" aria-describedby="popover-bottom">Popover on bottom with title and close button</button>
+        </div>
+    </div>
+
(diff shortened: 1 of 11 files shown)


sahil143:
https://rawgit.com/sahil143/patternfly-webcomponents/popover-dist/app/app.html?dir=pf-popover&file=index.html

src/pf-popover/index.html
```diff
@@ -0,0 +1,96 @@
...
+
+        <div id="div">
+            <pf-popover id="popover-right" placement="right" targetSelector="#btn-right-t-c" popoverTitle="Popver on right" dismissible="true">
+                popover text on right popover text on right popover text on right popover text on right popover text on right
+            </pf-popover>
+            <button id="btn-right-t-c" type="button" class="btn btn-primary" aria-describedby="popover-right">Popover on right with title and close button</button>
+        </div>
+
+        <div id="div">
+            <pf-popover id="popover-right" placement="right" targetSelector="#btn-right-t" popoverTitle="Popver on right">
```
priley86: Preference with attribute naming here is to use lower case w/ hyphen (i.e.: `popover-title` or just `title`). We could probably do the same w/ `target-selector` but that can be changed in a future PR... @cdcabrera thoughts on attribute naming?

src/pf-popover/index.html
```diff
@@ -0,0 +1,96 @@
...
+        </div>
+    </div>
+
+    <script src="//cdnjs.cloudflare.com/ajax/libs/webcomponentsjs/1.0.0/webcomponents-lite.js"></script>
+    <script src="//rawgit.com/webcomponents/custom-elements/master/src/native-shim.js"></script>
+    <script src="../../dist/js/patternfly.js"></script>
+    <script>
+        setTimeout(function () {
+            var pfPopoverRight = document.querySelector('[targetSelector="#btn-right"]');
+            pfPopoverRight.setInnerHtml("<span>Right content changed using public handler(setInnerHtml())!</span>");
```
priley86: Love it!!! This will add React/ng2/etc. integration. THANK YOU!!!

sahil143:
@priley86 made the changes regarding `pficon pficon-close` classes.

priley86:
@sahil143 i reviewed this w/ @cdcabrera some... let's try to add support for `isElementInViewport` like this: 
https://github.com/thednp/bootstrap.native/blob/master/lib/V3/popover-native.js#L118

We can add the same in `pf-tooltip` after 😸

src/pf-popover/index.html
cdcabrera: Prefer the hyphen approach since the majority of custom attributes we've come across also use it, such as ```data-custom```

src/pf-popover/pf-popover.component.js
```diff
@@ -0,0 +1,394 @@
...
+    const // link rect | window vertical and horizontal scroll
+      scroll = pfUtil.getScroll();
+
+    const //link real dimensions
+      linkDimensions = { w: rect.right - rect.left, h: rect.bottom - rect.top };
+
+    const popoverDimensions = { w: this.popover.offsetWidth, h: this.popover.offsetHeight };
+
+    //apply styling
+    if (/top/.test(this._placement)) { //TOP
```
cdcabrera: Seems like the perfect place for a ```switch(this._placement){}``` instead

sahil143:
@cdcabrera @priley86  I took a look at it and tried on smaller screen. Demo seems to be working fine(no tails). I think this only happens when we resize the browser window. What do you think?
a

cdcabrera:
@sahil143 the resize "bunch up" for tooltips does indeed happen on responsive screen-widths. If you notice your demo, while it's aimed at iPhone 6 resolution in the gif capture, it looks like a fixed width page, hence the "bunch up" won't fire

What we're curious about for the future is... 
- The behavior does emulate what's already implemented, and it appears the regular Bootstrap Popover handles the same way, but as extra what about an option/parameter/setting that lets the popover reposition itself automatically on smaller screens?
- and if the popovers are going to bunch up on small screens, is there a way to make the tail always align

src/pf-popover/pf-popover.component.js
```diff
@@ -0,0 +1,421 @@
...
+    this.popover = popovers[popovers.length - 1];
+    this.popover.style.display = 'block';
+    this.popover.setAttribute('class', `popover ${this._placement} ${this._animation}`);
+  }
+
+  /**
+   * update the placement of popover
+   */
+  _updatePlacement() {
+    return this._placement === 'top' ? 'bottom' : // top
```
priley86: would you mind simplifying this logic for us and removing the ternary? i.e. : 
```
if(position === top)
 return bottom;
if(position === bottom)
 return top;
if(position === left)
 return right;
if(position === right)
 return left;
return position;
```

src/pf-popover/pf-popover.component.js
```diff
@@ -0,0 +1,421 @@
...
+     * The popover open method
+     */
+  open() {
+    clearTimeout(this._timer);
+    this._timer = setTimeout(() => {
+      if (this.popover === null) {
+        this._createPopover();
+        this._stylePopover();
+        this._checkPlacement();
+        this._showPopover();
```
cdcabrera: Found a couple of more things... this sequence is causing multiple nodes to be appended, it's repeated around ```line... 405``` little bit of DOM pollution...
```javascript
_checkPlacement() {
  if (!pfUtil.isElementInViewport(this.popover)) {
    this._placement = this._updatePlacement();
    this._createPopover();
    this._stylePopover();
  }
}
```

Both sequences are load order specific... this isn't a requirement but if you can think of a way to use state that would push this component pretty far

src/pf-popover/pf-popover.component.js
```diff
@@ -0,0 +1,421 @@
...
+  }
+
+  /**
+   * check the placement of popover
+   */
+  _checkPlacement() {
+    if (!pfUtil.isElementInViewport(this.popover)) {
+      this._placement = this._updatePlacement();
+      this._createPopover();
+      this._stylePopover();
```
cdcabrera: Continuation... see comment above, related to causing multiple DOM nodes to be appended

src/pf-popover/pf-popover.component.js
```diff
@@ -0,0 +1,421 @@
...
+    }
+    //set popover content
+    popoverInner.innerHTML = this.content;
+
+    //append to the container
+    this._container.appendChild(clone);
+
+    //set reference to appended node
+    let popovers = this._container.querySelectorAll('.popover');
+    this.popover = popovers[popovers.length - 1];
```
cdcabrera: Not a requirement... Having trouble understanding if there will ever be more than one ```.popover``` element, but if it fits into the concept could try.... 

less specific way to handle this using ```:first-child```
```javascript
clone = clone.querySelector(':first-child');
this._container.appendChild(clone);
this.popover = clone;
```

src/pf-popover/pf-popover.component.js
priley86: Hey @sahil143, I noticed the same thing after discussing this w/ @cdcabrera. It appears that the `popover on left but not in view` example showcases this. It seems to be creating a `popover left fade` div unnecessarily (when the popover is actually on the right and inside another div). Let's try to correct the logic creating this `left` div in this scenario...

screen shot 2017-07-18 at 9 58 36 am

src/pf-popover/index.html
priley86: I've created [#53](https://github.com/patternfly-webcomponents/patternfly-webcomponents/issues/53) to track this. I think we can clean up attribute names to be consistent across in a new PR...

sahil143:
@cdcabrera made some changes to fix extra node and other minor changes. Please review when you when you get a chance. Thanks!!

sahil143:
@cdcabrera is there anything else we need on this?

cdcabrera:
@sahil143 changes looking good, taking a look at some of the coordinate logic. looks like we're a little off on the small screen, we're so close though. might be declaring it a bug/issue we can track against after we get it in. give you a heads up in a bit and then we can look at squashing those commits

sahil143:
@cdcabrera I noticed this too on small screen. If the position stays right then popover align with button with a long tail. Problem occurs when position is changed to the left because the element was not in view.
This also happens in Native javascript Bootstrap examples.

src/pf-popover/pf-popover.component.js
```diff
@@ -0,0 +1,430 @@
...
+  _stylePopover() {
+    const rect = this._target.getBoundingClientRect(); //popover real dimensions
+
+    const // link rect | window vertical and horizontal scroll
+      scroll = pfUtil.getScroll();
+
+    const //link real dimensions
+      linkDimensions = { w: rect.right - rect.left, h: rect.bottom - rect.top };
+
+    const popoverDimensions = { w: this.popover.offsetWidth, h: this.popover.offsetHeight };
```
cdcabrera: Apologizes for the delay... it looks like this call after the fact is affecting the positioning when the window is scaled... I tested it with something like this
```javascript
  //const popoverDimensions = { w: this.popover.offsetWidth, h: this.popover.offsetHeight };
  const popoverDimensions = { w: this._ogWidth, h: this._ogHeight };
```
where ```this._ogWidth``` and ```this._ogHeight``` are populated when the popover is first created, or where I used the original dimensions of the popover here instead of the scaled versions... 

So what's happening is...
- the popovers have a scaling min-width instead of being fixed... when the browser is scaled the popover scales width-wise hence grabbing the dimensions after the fact means the dimensions are scaled and the position is off. 

Since this potentially doesn't just affect popovers, but tooltips as well, i'll catch up with @bleathem and @priley86 and look at simply opening an issue on this after the fact. 

You can try to fix it if you want, but really for now just look at squashing those commits and we'll try to get this merged

sahil143:
@cdcabrera I squashed the commits. I will try to fix it until we are ready for Datepicker.

PR #65, opened 2017-08-06
Semantic release

dlabrecq:
Added semantic release to create a new package versions from master-dist.

Note that the patternfly-eng-release scripts will continue to run npm install, bower install, grunt build, grunt ngdocs:publish, npm test, nsp shrinkwrap audit, and will verify npm/bower installs. An npm shrinkwrap file will also be created and committed to master-dist along with other generated files. The scripts will continue to ensure master-dist is not published during pull requests and tagged builds -- it must be a merge to master and not a fork.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: node_js
 node_js:
   - '7'
+
 dist: trusty # needs Ubuntu Trusty
 sudo: false  # no need for virtualization.
 addons:
   chrome: stable # have Travis install chrome stable.
+
 env:
   global:
     - ENCRYPTION_LABEL: "d7ffc1eaa225"
@@ -14,7 +16,7 @@ env:
 
 before_install:
   - if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
-  - npm install git+https://github.com/patternfly/patternfly-eng-release.git
+  - npm install patternfly-eng-release
   - export CHROME_BIN=chromium-browser
   - export DISPLAY=:99.0
   - sh -e /etc/init.d/xvfb start
@@ -25,4 +27,11 @@ script:
   - sh -x ./node_modules/patternfly-eng-release/scripts/_build.sh -w
 
 after_success:
+  - 'if [[ "$TRAVIS_SECURE_ENV_VARS" = "true" && "$TRAVIS_BRANCH" = "master-dist" ]]; then
+       npm run semantic-release;
+     fi'
   - npm run publish-travis
+
+branches:
+  except:
+    - /^v\d+\.\d+\.\d+$/
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ https://patternfly-webcomponents.github.io/patternfly-webcomponents/
 ### Build
     npm install
     gulp build
+
 ### Serve
     gulp serve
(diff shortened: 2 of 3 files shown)


priley86:
thanks for setting this up @dlabrecq. I think this works fine for now here. I may revisit this a bit once we have Storybooks ready, but this gets us to a solid continuous release pipeline.

dlabrecq:
At some point, we should speak about Storybook more. I'm confused as how Storybook would effect this release? That, is other than how the showcase/demo is published using Brian's gh-publish script.

The patternfly-eng-release scripts currently push to master-dist mainly so we don't commit generated files to master. This keeps our PRs clean and relatively small -- it acts as a release branch.

Currently, the patternfly-webcomponents package includes generated CSS. Even if that was removed, we would still want to commit the dist directory which the npm release is based on.

Using patternfly-react as an example, I don't see CSS nor a dist directory in the master branch. The patternfly-react package has a transpiled lib directory, but that's only available if you install the package. Unfortunately, if I select the Github tag for release 0.5.0, the lib directory cannot be found -- only the source is available.

Whatever files are generated, I believe we want to commit them so they can be included with the tagged release on GitHub. Whether or not we push generated files to master Vs master-dist (a release branch) is another matter.

priley86:
@dlabrecq i'm fine w/ adopting the `master-dist` generated branch logic for maintaining Bower support (and that works nicely here). 

The only piece I see missing is what you've referenced around the the gh-publish script. The PF-React release script [here](https://github.com/patternfly/patternfly-react/blob/master/release.sh#L33) currently builds and pushes the updated storybook to the `gh-pages` branch. I'm hoping this can be maintained. The other subtle difference is the use of `commitizen` in pf-react. I'm also hoping to maintain this convention there (as that is a personal preference I think we're OK with there). Commitizen is of course optional, but for those using CLI it is definitely preferred.

I'm currently working to implement Storybooks from a Web Component perspective for this repo. Once you've adopted Storybooks in `patternfly-ng` maybe we can try to combine these release scripts for `gh-pages` deploys?

dlabrecq:
It's not my intention that patternfly-eng-release scripts publish to gh-pages. I'm leaving that to Brian's publish-ghpages.sh script -- that psrticular script is not part of patternfly-eng-release. 

If we want to replace Brian's script with Storybook, that would be fine. That will not effect the build or semantic release.

dlabrecq:
FYI, the semantic-release is working for patternfly, angular-patternfly, patternfly-ng, and patternfly-eng-release. I've updated this PR to mimic those repos.

Note that we will need to run npm publish manually and tag the master-dist branch as v0.0.10. At least for patternfly, angular-patternfly and patternfly-ng, that was required to get things working with the master-dist branch. The semantic-release expects the last released commit to be in the branch in order to determine the next version.

We can do that before or after this PR is merged. If we do it before, this merge will automatically increment the version number. If we that after the PR is merged, we will only see the version number incremented after a following commit.

I have no problem releasing and tagging master-dist, but wanted to give a heads up first.

priley86:
Sounds good! thanks for looking into this Dan!

dlabrecq:
Ok, I manually released v0.0.10 and tagged master-dist. This is ready to go!

PR #68, opened 2017-08-13
feat(notification-drawer): implement a Notification drawer web component

dabeng:
@priley86 @dlabrecq Would you like to review this PR again?

- [Demo under native js context](https://rawgit.com/dabeng/patternfly-webcomponents/notification-drawer-dist/src/pf-notification-drawer/index.html)

- [Using pf-notification-drawer in ReactJS](https://stackblitz.com/edit/using-pf-notification-drawer-in-reactjs)

### DIFF
--- a/src/pf-notification-drawer/pf-notification-drawer-accordion.component.js
+++ b/src/pf-notification-drawer/pf-notification-drawer-accordion.component.js
@@ -1,0 +1,51 @@
+import {pfUtil} from 'pf-utils.js';
+
+/*
+ * <b>&lt;pf-notification-drawer-accordion&gt;</b> element for Patternfly Web Components
+ *
+ * @example {@lang xml}
+ * <pf-notification-drawer-accordion>
+ *   <pf-notification-drawer-accordion-panel></pf-notification-drawer-accordion-panel>
+ *   <pf-notification-drawer-accordion-panel></pf-notification-drawer-accordion-panel>
+ *   . . .
+ * </pf-notification-drawer-accordion>
+ */
+
+export class PfNotificationDrawerAccordion extends HTMLElement {
+
+  /*
+   * An instance of the element is created or upgraded
+   */
+  constructor () {
+    super();
+  }
+
+  /*
+   * Called every time the element is inserted into the DOM
+   */
+  connectedCallback () {
+    let that = this;
+    pfUtil.addClass(this, 'panel-group');
+
+    let observer = new MutationObserver(function(mutations) {
+      if (that.querySelector('.drawer-pf-notification')) {
+        that.dispatchEvent(new CustomEvent('pf-notification-drawer.updategroups', { 'bubbles': true }));
+      }
+    });
+    observer.observe(this, { childList: true });
+  }
+
+  /*
+   * collapse the expanded panel
+   * @private
+   */
+  _collapseExpandedPanel (expanded) {
+    let originalExpanded = this.querySelector('.uncollapsed');
+    if (originalExpanded) {
+      pfUtil.removeClass(originalExpanded, 'uncollapsed');
+    }
+  }
+
+}
+
+window.customElements.define('pf-notification-drawer-accordion', PfNotificationDrawerAccordion);
--- a/src/pf-notification-drawer/pf-notification-drawer.scss
+++ b/src/pf-notification-drawer/pf-notification-drawer.scss
@@ -1,0 +1,35 @@
+pf-notification-drawer-accordion-panel {
+  display: block;
+  pf-notification-drawer-accordion-panel-collapse {
+    display: block;
+    position: relative;
+    height: 0px;
+    overflow: hidden;
+    -webkit-transition-property: height;
+    transition-property: height;
+    -webkit-transition-duration: 0.35s;
+    transition-duration: 0.35s;
+    -webkit-transition-timing-function: ease;
+    transition-timing-function: ease;
+    &.uncollapsed {
+      height: 326px;
+      &.empty-view {
+        height: 127px;
+        pf-notification-drawer-accordion-panel-body {
+          height: 0px;
+        }
+      }
+    }
+    pf-notification-drawer-accordion-panel-body {
+      display: block;
+      height: 284px;
+      overflow-y: auto;
+    }
+    .blank-slate-pf {
+      position: absolute;
+      top: 0px;
+      bottom: 1px;
+      width: 100%;
+    }
+  }
+}
--- a/src/pf-notification-drawer/pf-notification-drawer.spec.js
+++ b/src/pf-notification-drawer/pf-notification-drawer.spec.js
@@ -1,0 +1,141 @@
+describe ('PatternFly Notification Drawer Component Tests', function () {
+  let drawer, btn;
+
+  function render(html) {
+    let template = document.createElement('template');
+    template.innerHTML = html;
+    document.body.appendChild(document.importNode(template.content, true));
+  }
+
+  beforeEach(function (done) {
+    let mo = new MutationObserver(function () {
+      mo.disconnect();
+      done();
+    });
+    mo.observe(document.body, { childList: true });
+    render(
+      '<button id="btn-toggle-notification-drawer">Toggle Notification-Drawer</button>' +
+      '<pf-notification-drawer id="drawer" targetSelector="#btn-toggle-notification-drawer">' +
+      ' <pf-notification-drawer-header drawerTitle="Notes"></pf-notification-drawer-header>' +
+      ' <pf-notification-drawer-accordion>' +
+      '   <pf-notification-drawer-accordion-panel panelTitle="Tab 1">' +
+      '     <pf-notification-drawer-accordion-panel-collapse>' +
+      '       <pf-notification-drawer-accordion-panel-body>' +
+      '         <p class="drawer-pf-notification unread">' +
+      '           <span class="drawer-pf-notification-message">Notfication 1</span>' +
+      '         </p>' +
+      '         <p class="drawer-pf-notification">' +
+      '           <span class="drawer-pf-notification-message">Notfication 2</span>' +
+      '         </p>' +
+      '       </pf-notification-drawer-accordion-panel-body>' +
+      '     </pf-notification-drawer-accordion-panel-collapse>' +
+      '   </pf-notification-drawer-accordion-panel>' +
+      '  </pf-notification-drawer-accordion>' +
+      '</pf-notification-drawer>'
+    );
+    btn = document.querySelector('#btn-toggle-notification-drawer');
+    drawer = document.querySelector('#drawer');
+  });
+
+  afterEach(function () {
+    document.body.removeChild(drawer);
+    drawer = null;
+    document.body.removeChild(btn);
+    btn = null;
+  });
+
+  describe('Prperties testing', function () {
+    it('update the drawer title', function () {
+      drawer.querySelector('pf-notification-drawer-header').drawerTitle = 'Notices';
+      expect(drawer.querySelector('.drawer-pf-title').querySelector('h3').textContent).toBe('Notices');
+    });
+
+    it('update the panel title for one notification group', function () {
+      drawer.querySelector('pf-notification-drawer-accordion-panel').panelTitle = 'Group 1';
+      expect(drawer.querySelector('.panel-title').children[0].textContent).toBe('Group 1');
+    });
+  });
+
+  describe('Events testing', function () {
+    let callback = null;
+
+    beforeEach(function () {
+      callback = jasmine.createSpy();
+    });
+
+    it('pf-notification-drawer.markallread should be triggered when user clicks the "Mark All Read" button', function () {
+      drawer.addEventListener('pf-notification-drawer.markallread', callback);
+      document.querySelector('.mark-all-read').click();
+      expect(callback).toHaveBeenCalled();
+      expect(callback.calls.mostRecent().args[0].detail.source.panelTitle).toBe('Tab 1');
+      expect(drawer.querySelectorAll('.unread').length).toBe(0);
+      expect(drawer.querySelector('.panel-counter').textContent).toBe('0 New Events');
+      expect(btn.querySelector('.badge')).toBeNull();
+    });
+
+    it('pf-notification-drawer.clearall should be triggered when user clicks the "Clear All" button', function () {
+      drawer.addEventListener('pf-notification-drawer.clearall', callback);
+      document.querySelector('.clear-all').click();
+      expect(callback).toHaveBeenCalled();
+      expect(callback.calls.mostRecent().args[0].detail.source.panelTitle).toBe('Tab 1');
+      expect(drawer.querySelector('.panel-counter').textContent).toBe('0 New Events');
+      expect(btn.querySelector('.badge')).toBeNull();
+    });
+
+    describe('pf-notification-drawer.updateall testing', function () {
+      beforeEach(function (done) {
+        drawer.addEventListener('pf-notification-drawer.updateall', callback);
+        let mo = new MutationObserver(function () {
+          mo.disconnect();
+          done();
+        });
+        mo.observe(drawer.querySelector('.panel-body'), { childList: true });
+        drawer.querySelector('.panel-body').innerHTML =
+          '<p class="drawer-pf-notification">' +
+          '  <span class="drawer-pf-notification-message">Notfication A</span>' +
+          '</p>' +
+          '<p class="drawer-pf-notification unread">' +
+          '  <span class="drawer-pf-notification-message">Notfication B</span>' +
+          '</p>';
+      });
+
+      it('pf-notification-drawer.updateall should be triggered when user updates the data of one notification group', function () {
+        expect(callback).toHaveBeenCalled();
+        expect(callback.calls.mostRecent().args[0].detail.source.panelTitle).toBe('Tab 1');
+      });
+    });
+
+    describe('pf-notification-drawer.updategroups testing', function () {
+      beforeEach(function (done) {
+        drawer.addEventListener('pf-notification-drawer.updategroups', callback);
+        let mo = new MutationObserver(function () {
+          mo.disconnect();
+          done();
+        });
+        mo.observe(drawer.querySelector('.panel-group'), { childList: true });
+        drawer.querySelector('.panel-group').innerHTML =
+          '<pf-notification-drawer-accordion-panel panelTitle="New Tab">' +
+          '  <pf-notification-drawer-accordion-panel-collapse>' +
+          '    <pf-notification-drawer-accordion-panel-body>' +
+          '      <p class="drawer-pf-notification">' +
+          '        <span class="drawer-pf-notification-message">Notfication No.1</span>' +
+          '      </p>' +
+          '      <p class="drawer-pf-notification">' +
+          '        <span class="drawer-pf-notification-message">Notfication No.2</span>' +
+          '      </p>' +
+          '    </pf-notification-drawer-accordion-panel-body>' +
+          '  </pf-notification-drawer-accordion-panel-collapse>' +
+          '</pf-notification-drawer-accordion-panel>';
+      });
+
+      it('pf-notification-drawer.updategroups should be triggered when user updates the data of all the notification groups', function () {
+        expect(callback).toHaveBeenCalled();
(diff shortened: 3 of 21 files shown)


priley86:
a quick glance at this and I’m very impressed…great job incorporating the styles. Also nice work making the content easily editable and providing a React example. I would like to try to continue providing examples in Angular as well. Have you tried incorporating a WC into StackBlitz? This may be easier than using plnkr. I’m planning to do this with `pf-tabs` as well and ensure Angular support.
https://stackblitz.com/

@xudong as far as the code goes, one minor thing I would suggest going forward is to avoid the `forceUpdate` from the React side where possible and use the `MutationObserver` inside the WC to handle DOM updates. I think `forceUpdate` is an anti-pattern which we can use if you have to, but i think long term we will be able to use shadow dom slots…so to me, encapsulating this logic in a MutationObserver is a better pattern for now (that way we can make updates internally later without having to change the shim APIs too). thoughts?

I'll try to review this more thoroughly later this week...

src/pf-notification-drawer/pf-notification-drawer-accordion.component.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /*
+   * Called every time the element is inserted into the DOM
+   */
+  connectedCallback () {
+    let that = this;
+    pfUtil.addClass(this, 'panel-group');
+
+    let observer = new MutationObserver(function(mutations) {
```
dlabrecq: @priley86 I know you had reservations related to using MutationObserver and performance. Are we ok using MutationObserver?

src/pf-notification-drawer/pf-notification-drawer.scss
```diff
@@ -0,0 +1,46 @@
+.drawer-pf {
```
dlabrecq: In general, we should be using CSS Patternfly from Patterfly. 

For the Angular Patternfly repo; for example, we consume the Patternfly CSS. If end up creating custom CSS in the repo, that is typically a red flag and should be added to Patternfly instead. That way, the CSS can be consumed by other repos; Angular Patternfly, patternfly-ng, etc.

dlabrecq: This is a question regarding the overall architecture of patternfly-webcomponents. Why are we using SCCS instead of LESS? 

I don't believe we've added anything specific to SCCS? We should be consuming Patternfly LESS so we can take advantage of color variables for example. That is, if we must create custom CSS that's not included in Patternfly?

src/pf-notification-drawer/pf-notification-drawer.spec.js
```diff
@@ -0,0 +1,141 @@
...
+  let drawer, btn;
+
+  function render(html) {
+    let template = document.createElement('template');
+    template.innerHTML = html;
+    document.body.appendChild(document.importNode(template.content, true));
+  }
+
+  beforeEach(function (done) {
+    let mo = new MutationObserver(function () {
```
dlabrecq: Again, I'm not sure what our stand is regarding the use of MutationObserver. We seem to be using them in more than a few places.

src/pf-notification-drawer/pf-notification-drawer.spec.js
```diff
@@ -0,0 +1,141 @@
...
+      '         <p class="drawer-pf-notification">' +
+      '           <span class="drawer-pf-notification-message">Notfication 2</span>' +
+      '         </p>' +
+      '       </pf-notification-drawer-accordion-panel-body>' +
+      '     </pf-notification-drawer-accordion-panel-collapse>' +
+      '   </pf-notification-drawer-accordion-panel>' +
+      '  </pf-notification-drawer-accordion>' +
+      '</pf-notification-drawer>'
+    );
+    btn = document.querySelector('#btn');
```
dlabrecq: Is the target selector scoped?

What happens if there is another component in the page that uses a simple selector name like #btn? I'm thinking we should create names similar to our events to avoid possible collisions.

src/pf-utils/pf-utils.js
```diff
@@ -39,7 +43,7 @@ class PfUtil {
           return el;
         }
       } else if ( former === '.' ) {// If selector is a class
-        if ( new RegExp(latter).test(el.className) ) {
+        if ( this.hasClass(el, latter) ) {
```
dlabrecq: Why replace the RegExp?

Does this change break other components -- possibly relying on a regular expression?

dlabrecq (changes requested):
Looks good. I just have some questions and an issue with the amount of CSS we're adding here Vs adding that to Patternfly

src/pf-notification-drawer/pf-notification-drawer-accordion.component.js
priley86: @dlabrecq after observing the performance a bit more, I haven't noticed a huge impact. As @agathver was reporting before, a lot of Polymer still uses this pattern and the impact is minimal. In cases where transclusion is necessary (i.e. events/methods aren't as useful), I still see the MutationObserver as a viable option. This also helps us internally abstract the issue and minimize the impact later on when we adopt Shadow DOM. I.m.o. it will be much easier from an upgrade standpoint then. I should also add that I have heard many React/Angular developers recently discussing the addition of Shadow DOM into there solutions (so it'd be great if we can get going on this once it solidifies).

Another related topic, I am also now attempting to minimize DOM transclusion where possible (i.e. if i don't need to copy the nested DOM into a new template, I just let it render inside the CE, there's no need to create templates unnecessarily). I've proposed that DOM elements which are semantic `button`, `input`, etc. be left to the consumer (i.e. like our `pf-dropdown` does) and that non-semantic `divs` be encapsulated into CE where possible. This accomplishes three goals:
* simplifies the DOM structure / avoids unnecessary transclusion
* allow the outside framework to still add bindings/change inner DOM (if it's not transcluded)
* allows the CE to interact w/ the innerDOM and just provide minimal behaviors/enhance accessbility/ensure common styles/ etc.

Deciding which approach is easiest (MutationObserver, events/method updates, inner DOM) seems to just be a case by case analysis of what the consumer is comfortable with and is consistent. We'd definitely appreciate anymore feedback on this from an Angular perspective and I'll be trying to do some more Angular demos soon...

src/pf-notification-drawer/pf-notification-drawer.scss
priley86: +1 - if the CSS is *not* specific to the Custom Element or the approach here, it should be sourced from PF. Maybe we can contribute this particular CSS upstream?

@dlabrecq why SASS vs LESS? I don't really know... ;)

src/pf-notification-drawer/pf-notification-drawer.spec.js
priley86: I have no issue using them in cases they make sense.

src/pf-notification-drawer/pf-notification-drawer.scss
priley86: @srambach any thoughts on this CSS?

src/pf-notification-drawer/pf-notification-drawer.spec.js
priley86: +1

src/pf-notification-drawer/pf-notification-drawer.scss
dlabrecq: For OpenShift IO, we converted all their SCCS to LESS. Mainly so they could build their own CSS along with Patternfly CSS and patternfly-ng. Having patternfly-webcomponents be SCCS while Patternfly is LESS may end up being a show stopper for some? That is, considering they will both be compiled together in apps. At least, OSIO could not consume this as is.

priley86: Ok, noted in a new issue [#71](https://github.com/patternfly-webcomponents/patternfly-webcomponents/issues/71). I'm fine w/ moving to LESS.

src/pf-utils/pf-utils.js
dabeng: Because RegExp writing can not distinguish class "aa" and class "aa-bb"

dabeng: Because RegExp writing can not distinguish class "aa" and class "aa-bb".

src/pf-notification-drawer/pf-notification-drawer.scss
dabeng: I think adding custom CSS for WCs of patternfly-webcomponents isn't unavoidable only if the structure of component is simple enough. On the other hand, Pantternfly is under no obligation to include all the css codes for derivatives like patternfly-react or patterfnly-ng. Maybe we can replace ".drawer-pf" with "pf-notification-drawer" using wc tag name to replace classname within patternfly.

dabeng:
@priley86 , I believe that shaw dom adoption will have a serious impact on flexibility and customizability of WCs. IMO, it‘s a chicken ribs inner web components skill-sets.

dabeng:
Hi @priley86 , I have added function updateObjectInArray(). forceUpdate() is abandoned. :)

dabeng:
@priley86 , this is new demo on Stackblitz.io

src/pf-notification-drawer/pf-notification-drawer.scss
dlabrecq: My point is there should be no CSS derivatives for patternfly-react, patternfly-ng, etc. The CSS and HTML layout should be the same for any implementation. 

The idea is that we should be able to create an example of this in the Patternfly Core test pages. Therefore, the CSS really belongs there so consumers of Patternfly Core can use it -- as well as patternfly-react, patternfly-ng, etc.

Ideally, there should be no CSS in this repo at all. Similar to Amgular Patternfly, there may be cases where we need a small amount of CSS, but it was added with the intention that it's back ported to Patternfly.

priley86:
@dabeng we have been discussing the same thing. The Shadow DOM will have a dramatic impact and I hope we can adopt it sooner than later...

priley86:
@dabeng some minor feedback on the stackblitz...can you possibly split out any of the code into separate templates/separate files? it's a bit hard to read with so much in one file...

src/pf-notification-drawer/pf-notification-drawer.spec.js
dabeng: In fact, we have the following code snippets to avoid collisions.
```js
  afterEach(function () {
    document.body.removeChild(drawer);
    drawer = null;
    document.body.removeChild(btn);
    btn = null;
  });
```
While, appending specific id will becoming the belt and braces. Very nice :)

dabeng:
Hi @priley86 , I have slimmed down the pf-notification-drawer demo on StackBlitz.

src/pf-notification-drawer/pf-notification-drawer.scss
dabeng: FYI, patternfly is based on bootstrap v3. Accordion component is ready-made for notification-drawer of patternfly. However, in patternfly-webcomponents, I have to create accordion from the scratch and append a few specific css code. 😢

priley86:
@dabeng i have had a bit more time to review this... here's my thoughts so far:

* in general, this is awesome! I love the StackBlitz example and this API works very well for React. I see it as a natural fit in the React ecosystem. While I know it will take more time, I'd love to see an Angular2 example on StackBlitz as well. I don't think that's a requirement in getting this merged, but a nice addition to really showcase this even further...
* it would be nice to provide some class override support for the following elements. allowing the consumer to modify the class of the first child in the template let's them further customize when needed. There is example of watching the `class` element in pf-tabs [here](https://github.com/patternfly-webcomponents/patternfly-webcomponents/blob/master/src/pf-tabs/pf-tabs.component.js#L59). Basically we are just watching the class attribute and overriding the first child when the consumer provides a class...
   * `pf-notification-drawer`, `pf-notification-drawer-accordion`, `pf-notification-drawer-header`, `pf-notification-drawer-accordion-panel`, `pf-notification-drawer-accordion-panel`
* it would be nice to provide `pf-notification-drawer*.initialized` events after the `connectedCallback` has initialized. We're doing this elsewhere so it would be nice to stay consistent and give the consumer the option to run additional logic after initialization is thru.
* Regarding the additional CSS:
   * In general, i agree that we should be doing our utmost to consume upstream CSS as is. However, there are cases in our downstream repos (React, Angular, Web Components) where the DOM structure is changed slightly causing some additional CSS. You've done a good job keeping this to a minimum here though, so my feedback would be the following:
   * prefer to style the **custom element** at the root of the selector instead of creating a new class. i.e. the root selector should be: `pf-notification-drawer {` instead of `.drawer-pf {`. This quickly indicates that we are styling the custom element in this case (and is consistent with our other minor style enhancements in this project to date). Ideally though, in the long term these should be contributed upstream (so we should keep them as minimal as possible). I think we are quite a ways out from incorporating custom elements into the official PF styles, but we can still minimize these where possible. We should be regularly reporting these to the CSS team and trying to enhance the upstream CSS where possible to avoid this when we can, but it should not block us from continuing to write custom elements.

Lastly, I noted you used an `html-import` in the test page. This is fine in a test page, but just want to clarify for everyone else we cannot use these in Angular/React apps (and prefer ES6 modules like you have used in StackBlitz).

Thanks again for the hard work on this, it's really showing!!!

src/pf-notification-drawer/pf-notification-drawer.scss
priley86: @dabeng @dlabrecq I wrote a few more comments about the CSS below...let me know what you guys think?

dabeng:
Hi @priley86 , I has committed new patch and updated the demo on Stackblitz.
1. Now the pf-notification-drawer.scss is shown as below:
```css
pf-notification-drawer-accordion-panel {
  display: block;
  pf-notification-drawer-accordion-panel-collapse {
    display: block;
    ...
    &.uncollapsed {
      height: 326px;
      &.empty-view {
        height: 127px;
        .pf-notification-drawer-accordion-panel-body {
          height: 0px;
        }
      }
    }
    pf-notification-drawer-accordion-panel-body {
      display: block;
      height: 284px;
      overflow-y: auto;
    }
    .blank-slate-pf {
      position: absolute;
      top: 0px;
      bottom: 1px;
      width: 100%;
    }
  }
}
```
2. I added event "pf-notification-drawer.initialized" to the pf-notification-drawer;
3. I want to keep the using of "html imports" in order to show users of patternfly-webcomponents another aspect of web components skills-set;
4. You said that we should allow users to insert their custom classname into internal structure of pf-notification. I'm sure it's awesome feature, but give me some to reconsider whether we can only append custom classname to the first child node :blush:

priley86:
@dabeng for #1,#2, and #3 above, I'm happy with! Thanks very much for making these changes.

For #4 (regarding class names), after considering this further and discussing with you, I think we should be using named attributes on the CE to override the classes of its inner dom. Initially I just proposed calling it "class" because it is easier to render a non-custom attribute in React, however this approach won't scale (say for example, we need to override multiple classes on different children). So i'd propose we do something like the following to extend flexibility on the styling of these inner elements

```
<pf-notification-drawer-accordion-panel panel-class="panel panel-default my-class">
```
which then renders a the "panel-class" to the panel div appropriately:
```
<div class="panel panel-default my-class">
```

I took a similar approach when recently updating #67 and you can see how I was able to render the custom attributes in React using the split operator (in the stackblitz example).

After reviewing the approach here and reimplementing the `pf-tabs` API, I'll be following up shortly w/ some additional concerns and challenges we might have here w/ transcluded DOM. 

Thanks again for all the hard work on this!

dabeng:
Hi @priley86 , I tried to introduce the shadow dom **v1** into pf-notification-drawer in the last two days. This is elementary demo -- https://rawgit.com/dabeng/patternfly-webcomponents/nd-shadow-dist/src/pf-notification-drawer/index.html. I found that shadow dom can simplify the custom elements' internal structure. This is great profit for us. The following is compare results.
**without shadow dom**
```html
<pf-notification-drawer targetSelector="#btn-toggle-notification-drawer">
        <pf-notification-drawer-header drawerTitle="Notifications Drawer"></pf-notification-drawer-header>
        <pf-notification-drawer-accordion>
          <pf-notification-drawer-accordion-panel panelTitle="Notification Tab 1">
          <pf-notification-drawer-accordion-panel-collapse>
            <pf-notification-drawer-accordion-panel-body></pf-notification-drawer-accordion-panel-body>
          </pf-notification-drawer-accordion-panel-collapse>
          </pf-notification-drawer-accordion-panel>
          <pf-notification-drawer-accordion-panel panelTitle="Notification Tab 2">
            <pf-notification-drawer-accordion-panel-collapse>
              <pf-notification-drawer-accordion-panel-body></pf-notification-drawer-accordion-panel-body>
            </pf-notification-drawer-accordion-panel-collapse>
          </pf-notification-drawer-accordion-panel>
          <pf-notification-drawer-accordion-panel panelTitle="Notification Tab 3">
            <pf-notification-drawer-accordion-panel-collapse>
              <pf-notification-drawer-accordion-panel-body></pf-notification-drawer-accordion-panel-body>
            </pf-notification-drawer-accordion-panel-collapse>
          </pf-notification-drawer-accordion-panel>
        </pf-notification-drawer-accordion>
      </pf-notification-drawer>
```
**with shadow dom**
```html
      <pf-notification-drawer targetSelector="#btn-toggle-notification-drawer" drawerTitle="Notifications Drawer">
        <pf-notification-drawer-panel panelTitle="Notification Tab 1" slot="panel"></pf-notification-drawer-panel>
        <pf-notification-drawer-panel panelTitle="Notification Tab 2" slot="panel"></pf-notification-drawer-panel>
        <pf-notification-drawer-panel panelTitle="Notification Tab 3" slot="panel"></pf-notification-drawer-panel>
      </pf-notification-drawer>
```

On the other hand, shadow dom causes a lot of pain for users to reuse existing style sheets for web components. I really don't find any other ways to style the elements within shadow root except for copying all the related css code into shadow dom.

dabeng:
Hi @priley86 , could you give me the explicit advice, What should I commit next step? Shadow dom or custom-class

mcarrano:
@dabeng I took a look at your test page from a design perspective.  For the most part, this looks great.  There was only one issue I noticed.  When opening any of the accordion panels, the caret at the left of the title should point down and point to the right when the panel is closed.  In your implementation I notice that the caret always points right.

It looks like there is actually a bug on the design page where this is inconsistent.  But if you match the HTML/CSS implementation here, that is the correct behavior: http://www.patternfly.org/pattern-library/communication/notification-drawer/#/code

dabeng:
Sorry @mcarrano , It's my mistake. The bug you mentioned above I have fixed already, but I forgot to update the demo link 😞 . Please review https://rawgit.com/dabeng/patternfly-webcomponents/notification-drawer-dist/src/pf-notification-drawer/index.html 😊

mcarrano:
Looks good, @dabeng Thanks for the update.

dabeng:
Hi @dlabrecq , I think this PR has implemented every details described within task PTNFLY-2445. Shall we merge it now?

Additionally, whether introducing shadow dom or skate.js, it's the matters beyond the scope of the user story PTNFLY-2445. We should create another issues to track it.

dlabrecq:
I'll have to defer to @priley86 here as I'm not following the efforts around Shadow DOM. It's not clear if the team decided to introduce it or not? I have some reservations about using that after some issues experienced with Angular apps.

dabeng:
Hi @priley86 ,this PR has been pending more than one week, but I think this PR has implemented every details described within task PTNFLY-2445. Shall we merge it now and create another task to track skate.js introduction ?

priley86:
hey @dabeng, i'm fine with merging this PR and closing out PTNFLY-2445. I think our discoveries in this work have definitely warranted a move to Shadow DOM for complex components such as notification drawer.

@dlabrecq @dabeng I should have a functioning prototype which builds some of our current web components alongside a React Storybook and our Patternfly NG template early next week. It also adds a simple Skate JS component to the mix. I'd like to share this next week and start iterating in a new repo soon under the Patternfly Org. I'd also like to continue our Shadow DOM implementations there. It will be quite easy to port over the components here (that we'd like to keep), we'll just need to add some Storybook/NG demos for each component. 
 
@dabeng feel free to continue exploring your Shadow DOM implementations with this component as you have time in future stories. I definitely have some ideas w.r.t. to Shadow DOM I'd like to discuss with you guys. We should have a pretty solid foundation to start testing Shadow DOM very soon...

